PR #23, opened 2020-09-28
Support Node 10

yoannmoinet:
### What and why?

Using the build plugin in Yarn's pipeline revealed that the build-plugin doesn't support Node 10.x because of the use of `Array.prototype.flat()`.

### How?

- Use an helper function instead of the native function.
- Run tests on older Node versions.


arcanis (approved):
LGTM

PR #24, opened 2020-10-28
More tags and more tests

yoannmoinet:
### What and why?

We needed to have tags on every `modules`, `chunks` and `assets` to track down from which entry/chunk they were coming from.

### How?

So `assets` and `modules` get two new tags for each `entry` and `chunk` that is "generating" it.
For `chunks`, they get tagged with their `entry`.

Since it's touching one of the core part of the plugin, we needed to have good testing and implemented most of the infra to allow testing on multiple version of webpack.

So we have two mocked projects, one with webpack 4 and another one with webpack 5, hence the very large dependency addition.

We also have improved performance in the aggregator by using an indexed object of values.


src/hooks/datadog/aggregator.ts
```diff
@@ -183,98 +184,221 @@ const findDependencies = (
...
+    indexed: IndexedObject,
+    context: string
+): Metric[] => {
     return flattened(
-        clonedModules.map((module) => {
-            // Modules are sometimes registered with their loader.
-            if (module.name.includes('!')) {
-                return [];
+        Object.values(indexed.modulesPerName).map((module) => {
+            const chunks = getChunksFromModule(stats, indexed.chunksPerId, module);
```
Youpinadi, line 272: nit: split the body of the map function into a separate function

Youpinadi (approved):

src/hooks/datadog/aggregator.ts
```diff
@@ -14,14 +14,15 @@ import {
...
+    Entry,
+    IndexedObject,
 } from '../../types';
 import { getMetric } from './helpers';
 import { Metric, MetricToSend, GetMetricsOptions } from './types';
 
 const flattened = (arr: any[]) => [].concat(...arr);
 
-const getType = (name: string) => name.split('.').pop();
+const getType = (name: string) => (name.indexOf('.') >= 0 ? name.split('.').pop() : 'unknown');
```
arcanis, line 25: Nit:
```suggestion
const getType = (name: string) => (name.includes('.') ? name.split('.').pop() : 'unknown');
```

arcanis (approved):

PR #28, opened 2021-05-28
Add chunkNames to modules output

yoannmoinet:
### What and why?

We're missing chunk names in the modules report.

### How?

Add `chunkNames` information for each module in the report.
Sometime, a module can be in multiple chunks, so it needs to be an array.


Miz85 (approved):
👍🏼  Thanks!

PR #29, opened 2021-08-05
Allow object in output option

yoannmoinet:
### What and why?

It can take some time to write files depending on their size.
And we don't need all of them.

### How?

Add the possibility to have an object in the `output` option.

```javascript
{
    output: {
        destination: 'path/to/destination/',
        timings: true
    }
}
```


hinosxz (approved):
LGTM!

jonathankauch (approved):

PR #36, opened 2021-10-13
Esbuild support

yoannmoinet:
### What and why?

Add esbuild support.

### How?

- Refactor a bunch of stuff to have multiple bundlers co-existing and re-using the same code as much as possible.
- Add the Esbuild plugin.

# Breaking changes

- no more `modules.tree.size` or `modules.tree.count` as it was kind of useless, complicated and noisy.
- `output.stats` becomes `output.bundler` to cover bundler's specific reports.
- `stats.json` becomes `bundler.json` to cover multiple bundlers.

# TODO

- [x] Esbuild plugin
- [x] Print Esbuild top stats
- [x] Write Esbuild file reports
- [x] Merge bundlers metrics processing
- [x] Update documentation
- [x] Esbuild module metrics
- [x] Esbuild assets metrics
- [x] Esbuild entries metrics
- [x] Add testing


bin/oss/index.js
```diff
@@ -128,11 +141,17 @@ class OSS extends Command {
...
+    })
+);
+OSS.addOption(
+    `directories`,
+    Command.Array(`-d,--directories`, {
+        description: 'On which directories to add the Open Source header?',
+    })
+);
 
 module.exports = [OSS];
```
yoannmoinet, line 157: Just some linting.

src/webpack/tapables.ts
```diff
@@ -46,39 +39,45 @@ export class Tapables {
...
+        timing.increment += 1;
+        this.timings.set(pluginName, timing);
     }
 
     getResults(): TapablesResult {
         const timings = this.timings;
 
         // Aggregate the durations for each plugin.
-        for (const [tapableName, tapable] of Object.entries(this.timings)) {
+        for (const [tapableName, tapable] of this.timings.entries()) {
```
arcanis, line 71: iirc you don't need to specify `.entries()` when iterating over standard Maps.

src/esbuild/index.ts
```diff
@@ -0,0 +1,64 @@
...
+                start: startBuild,
+                report: {
+                    timings: {
+                        tapables: plugins,
+                        modules,
+                    },
+                    dependencies: moduleResults,
+                },
+                bundler: {
+                    esbuild: {
```
arcanis, line 41: Does it mean the metrics will be uploaded in separate "namespaces"? How will that interact with things like the budget plan?

src/__tests__/webpack/webpack.test.ts
```diff
@@ -2,8 +2,8 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2019-Present Datadog, Inc.
 
-import { BuildPlugin } from '../webpack';
-import { mockCompiler } from './helpers/testHelpers';
+import { BuildPlugin } from '../../webpack';
+import { mockCompiler } from '../helpers/testHelpers';
```
arcanis, line 6: Are the esbuild tests planned for the same PR?

src/esbuild/plugins.ts
```diff
@@ -0,0 +1,112 @@
...
+    build: PluginBuild,
+    pluginName: string
+): PluginBuild => {
+    const newBuildObject: any = Object.assign({}, build);
+    for (const fn of Object.values(FN_TO_WRAP)) {
+        newBuildObject[fn] = async (opts: any, cb: any) => {
+            // TODO Remove debug.
+            if (pluginsMap.get(pluginName)) {
+                console.log(`Already got the timing for ${pluginName}`, pluginsMap.get(pluginName));
+            }
```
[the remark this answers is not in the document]
yoannmoinet: Ah, good catch, will update them.

src/esbuild/index.ts
yoannmoinet: Nope, this doesn't change the metric's namespace.
The metric's namespace is decided at the integration level, in the settings.
Right now, we have it setup as `dd.web_ui.webpack` for our webpack build.

https://github.com/DataDog/web-ui/blob/8d8e8bed1a0c4dcca1a9da42846c94390e0b91a5/webpack.config.js#L513

Here, it's just for better handling esbuild's specific data internally.

That being said, I think we should use a different namespace for esbuild.
It will ease the migration for the budget plan too.
Otherwise, during the migration, we may end up with metrics overwriting each others (between webpack and esbuild builds) and the budget will go awry.

src/__tests__/webpack/webpack.test.ts
yoannmoinet: Yes, I'm working on them.

src/hooks/datadog/metrics/webpack.ts
```diff
@@ -0,0 +1,279 @@
...
+    );
+};
+
+export const getAssets = (stats: StatsJson, indexed: WebpackIndexedObject): Metric[] => {
+    const assets = stats.assets;
+    return assets.map((asset) => {
+        const chunks = asset.chunks.map((c) => indexed.chunksPerId[c]);
+        const entries: Set<string> = new Set();
+        for (const chunk of chunks) {
+            getEntriesFromChunk(chunk, indexed, entries);
```
arcanis, line 162: nit: it's a little surprising to have a `getSomething` function with exclusively side effects

arcanis (approved):
Didn't spot obvious problems; it could benefit from more tests, but iirc you already are working on that 👍

src/hooks/datadog/metrics/webpack.ts
yoannmoinet: Is it about the `entries` parameter?
I think it can be removed, as it's filled in by default.

README.md
```diff
@@ -71,6 +75,23 @@ module.exports = {
...
+
+```js
+const esbuild = require('esbuild');
+const { BuildPlugin } = require('@datadog/build-plugin/dist/esbuild');
+
+esbuild.build({
+  [...] // All your configuration needs
+  plugins: [
+    [...] // All your plugins
+    BuildPlugin()
```
arcanis, line 90: Position doesn't matter for this one (unlike Webpack where the plugin has to be last)?

Btw, does it have an overhead at build time?

package.json
```diff
@@ -28,7 +28,9 @@
     },
     "dependencies": {
         "chalk": "2.3.1",
+        "esbuild": "0.12.26",
```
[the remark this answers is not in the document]
jakub-g: I see @yoannmoinet you added a peer dep so this could be removed I guess?

README.md
yoannmoinet: Yes, the position doesn't matter here because esbuild runs the plugins' setup once they all registered and not as soon as they instantiate.

The overhead at build time is as minimal as for webpack.

package.json
```diff
@@ -41,6 +43,7 @@
         "@types/fs-extra": "8.1.0",
         "@types/jest": "26.0.0",
         "@types/node": "13.13.4",
+        "@types/pretty-bytes": "5.2.0",
```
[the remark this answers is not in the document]
yoannmoinet: Good catch, removed it.

package.json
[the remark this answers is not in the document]
yoannmoinet: If I remove it from here, I can't import it anymore.
Not sure exactly how peer deps are actually working.

PR #39, opened 2022-04-21
Share dashboard's json

### LINKED ISSUE #26: [feature request] Default Dashboard JSON Configuration
It would be useful to be able to have access to the dashboard JSON required to generate this picture, rather creating the tiles individually for each new org we set this up for

dash

 I've just set this plugin up for nycvaccinelist.com, and wanted to check if there's a way we can get this default integration dashboard in a few clicks.

yoannmoinet:
### What and why?

We have a nifty little dashboard that could be shared.

### How?

- Add the `dashboard.json` template.
- Add the command `yarn cli dashboard -p <your.prefix>` to output that configuration with a custom prefix.

fixes #26


SBoudrias (approved):

hydrosquall (approved):
The Prefix CLI option is a nice touch! Thank you.

bin/dashboard/index.js
```diff
@@ -0,0 +1,29 @@
...
+
+const JSON_PATH = path.join(__dirname, '../../assets/dashboard.json');
+
+class Dashboard extends Command {
+    async execute() {
+        if (!this.prefix) {
+            throw new UsageError('Missing --prefix option.');
+        }
+        const content = await fs.readFile(JSON_PATH, 'utf-8');
+        console.log(content.replace(/\{\{PREFIX\}\}/g, this.prefix));
```
hydrosquall, line 17: Seems reasonable to do string substitution since there's only one value. If the default gets more elaborate, we might consider treating the JSON file as a template literal instead.

bin/dashboard/index.js
```diff
@@ -0,0 +1,29 @@
...
+const { Command, UsageError } = require(`clipanion`);
+const fs = require('fs-extra');
+const path = require('path');
+
+const JSON_PATH = path.join(__dirname, '../../assets/dashboard.json');
+
+class Dashboard extends Command {
+    async execute() {
+        if (!this.prefix) {
+            throw new UsageError('Missing --prefix option.');
```
hydrosquall, line 14: If we are requiring the prefix for the generated dashboard inside the datadog , I wonder if we should make it a required parameter. Currently this might not work if someone uses the default prefix of `""` since their metric names would begin with a `.`

https://github.com/DataDog/build-plugin/blob/5d437ddef97b3382605f156d0af23c92bab988c3/src/hooks/datadog/README.md#datadogprefix

README.md
```diff
@@ -141,6 +141,8 @@ An object used to automatically send your build data to Datadog.
 
 ![](./assets/datadog-dashboard.png)
 
+> You can get this dashboard's configuration by running `yarn cli dashboard --prefix <your.prefix>`.
+
 The most basic configuration looks like this, consult
```
hydrosquall, line 146: We might want to provide a default prefix inside the basic configuration below if the default of `""` doesn't create the intended results.

bin/dashboard/index.js
yoannmoinet: We'll cross that bridge once we get there.

README.md
yoannmoinet: That will be done, if needed, in a different PR.

bin/dashboard/index.js
yoannmoinet: Good point.
I'll make it accept no prefix.

PR #48, opened 2023-09-13
Fix loaders reporting

yoannmoinet:
### What and why?

Loaders' metrics were not reported as expected.

### How?

We were using a `Map` just like an Object/Array.


Ayc0 (approved):
🔥

PR #50, opened 2023-09-13
Update yarn

yoannmoinet:
### What and why?

2.0.0 is old.

### How?

3.2.4 is better.
But it breaks the `licenses-csv` plugin.
To be fixed in a different PR.


Ayc0 (approved):

yoannmoinet:
Fixed in https://github.com/DataDog/build-plugin/pull/49

PR #54, opened 2023-12-12
Node: bump to 18 & 20

Ayc0:
### What and why?

Node 12 that we were using is long deprecated. Same for node 10 & 14 used in the CI.
The current LTS are 18 & 20:

image

### How?

- Update node everywhere
- Enable `--openssl-legacy-provider` because:
  image


yoannmoinet (approved):
Thanks

PR #55, opened 2023-12-12
Upgrade yarn

Ayc0:
### What and why?

We're currently using `yarn` v2 rc which is outdated and buggy.
Let's update to the v4.

Also our custom plugin to get all OSS licenses is broken (already on yarn v2), let's update it too

### How?

- Upgrade yarn to v4 & run `yarn install`
- Re-write `yarn oss` by using https://github.com/mhassan1/yarn-plugin-licenses & a custom script around it.
  To note: this plugin doesn't have access to the emails of the maintainers, so we lost this in the csv 😕 
  image
- Update yarn config to install packages for both darwin & linux


bin/oss/index.js
```diff
@@ -96,14 +96,59 @@ class OSS extends Command {
...
         } catch (e) {
             // eslint-disable-next-line no-console
             console.log(e);
         }
 
-        if (stdout) {
-            return fs.writeFile(path.join(ROOT, 'LICENSES-3rdparty.csv'), stdout);
+        const licenses = new Map();
+
+        const nameRegex = /^(@[.*?]\/.*?|[^@]+)@(.+?):.+?$/;
```
yoannmoinet, line 107: Can you add a comment explaining what it does with some examples please.

LICENSES-3rdparty.csv
```diff
@@ -1,288 +1,312 @@
...
+esprima,npm,BSD-2-Clause,Ariya Hidayat (http://esprima.org)
+esquery,npm,BSD-3-Clause,Joel Feenstra (https://github.com/estools/esquery/)
+esrecurse,npm,BSD-2-Clause,(https://github.com/estools/esrecurse)
+estraverse,npm,BSD-2-Clause,(https://github.com/estools/estraverse)
+esutils,npm,BSD-2-Clause,(https://github.com/estools/esutils)
+events,npm,MIT,Irakli Gozalishvili (http://jeditoolkit.com)
+evp_bytestokey,npm,MIT,Calvin Metcalf (https://github.com/crypto-browserify/EVP_BytesToKey)
+exec-sh,npm,MIT,Aleksandr Tsertkov
+execa,npm,MIT,Sindre Sorhus (sindresorhus.com)
+exit,npm,MIT,"Cowboy" Ben Alman (https://github.com/cowboy/node-exit)
```
yoannmoinet, line 188: The [parsing](https://github.com/DataDog/build-plugin/blob/87f850209a4c1a2630e9c31eda07f069f08d3cee/LICENSES-3rdparty.csv) seems to fail on this line, maybe because of `"`.

yoannmoinet (commented):
Mostly LGTM, just some nits.

bin/oss/index.js
Ayc0: I added a comment, please tell me what you think about it 👌

yoannmoinet (approved):

PR #61, opened 2024-03-20
[v2.0.0] Re design of the plugin

yoannmoinet:


jakub-g (approved):
Approving to unblock

packages/assets/package.json
```diff
@@ -0,0 +1,13 @@
+{
+    "name": "@datadog/build-plugins-assets",
+    "packageManager": "yarn@4.0.2",
+    "license": "MIT",
+    "private": true,
+    "author": "Datadog",
+    "exports": {
```
Ayc0, line 7: FYI (not blocking): instead of using `exports` for such static files, you can use `files` to mark the `src/` folder as public
Or even don't write anything, and keep all the files top level in this package

packages/core/src/hooks/datadog/metrics/esbuild.ts
```diff
@@ -42,7 +42,8 @@ export const getIndexed = (stats: EsbuildStats, context: string): EsbuildIndexed
     if (Array.isArray(stats.entrypoints)) {
         // We don't have an indexed object as entry, so we can't get an entry name from it.
         for (const entry of stats.entrypoints) {
-            const realEntry = getModulePath(entry, context);
+            const fullPath = typeof entry === 'object' ? entry.in : entry;
```
Ayc0, line 45: Not blocking: I didn't check the type, but if `entry` is `null`, `typeof entry` will also say `'object'`
If you want to handle such cases, you need to do `entry && typeof entry === 'object'`

packages/core/src/hooks/outputFiles.ts
```diff
@@ -5,19 +5,16 @@
...
 
 type FilesToWrite = {
     [key in Files]?: { content: any };
 };
 
-const output = async function output(
-    this: BuildPlugin,
-    { report, metrics, bundler }: HooksContext
-) {
+const output = async function output(this: BaseClass, { report, metrics, bundler }: HooksContext) {
```
Ayc0, line 17: Note: no need for `const output =`, you're already defining a function named `output`
```suggestion
async function output(this: BaseClass, { report, metrics, bundler }: HooksContext) {
```

packages/core/src/BaseClass.ts
```diff
@@ -73,7 +71,7 @@ export class BaseClass {
     // Will apply hooks for prehookName, hookName and posthookName
     async applyHooks(hookName: HOOKS) {
         const applyHook = (name: WRAPPED_HOOKS) => {
-            const proms = [];
+            const proms: Promise<any>[] = [];
```
Ayc0, line 74: Note: instead of `any` (if it matters), you can either use a generic, or `unknown`

packages/core/README.md
```diff
@@ -0,0 +1 @@
+# Core
```
Ayc0, line 1: You can remove this file as it doesn't add that much value

packages/core/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "license": "MIT",
+    "private": true,
+    "author": "Datadog",
+    "main": "./src/index.js",
+    "exports": {
+        "./*": {
+            "import": "./src/*.ts",
+            "require": "./src/*.ts"
+        },
+        "./hooks/datadog": {
```
Ayc0, line 13: 2 things:
1. I think this needs to go before the `*`
2. if you're using the same for `import` & `require`, you can remove both, and just do `"./hooks/datadog": "./src/hooks/datadog/index.ts"` (same for `./*`)

packages/esbuild-plugin/src/plugins.ts
```diff
@@ -7,15 +7,10 @@
...
+import { TimingsMap, Timing, Value } from '@datadog/build-plugins-core/types';
+import { getContext, formatModuleName } from '@datadog/build-plugins-core/helpers';
 
-enum FN_TO_WRAP {
-    START = 'onStart',
-    LOAD = 'onLoad',
-    RESOLVE = 'onResolve',
-    END = 'onEnd',
-}
+const FN_TO_WRAP = ['onStart', 'onLoad', 'onResolve', 'onEnd'];
```
Ayc0, line 13: If you want the type of `FN_TO_WRAP` to not be `string[]` but constant, and not have to do `fn as keyof PluginBuild`, you can just do:
```suggestion
const FN_TO_WRAP = ['onStart', 'onLoad', 'onResolve', 'onEnd'] as const;
```

packages/esbuild-plugin/src/plugins.ts
```diff
@@ -46,11 +41,11 @@ export const wrapPlugins = (build: PluginBuild, context: string) => {
...
     build: PluginBuild,
     pluginName: string,
-    context: string
+    context: string,
 ): PluginBuild => {
     const newBuildObject: any = Object.assign({}, build);
-    for (const fn of Object.values(FN_TO_WRAP)) {
-        newBuildObject[fn] = async (opts: any, cb: any) => {
+    for (const fn of FN_TO_WRAP) {
+        newBuildObject[fn as keyof PluginBuild] = async (opts: any, cb: any) => {
```
Ayc0, line 48: See my other comment, you should be able to remove this `as`
```suggestion
        newBuildObject[fn] = async (opts: any, cb: any) => {
```

packages/esbuild-plugin/src/plugins.ts
```diff
@@ -62,8 +57,8 @@ const getNewBuildObject = (
                 name: fn,
                 values: [],
             };
-
-            return build[fn](opts, async (...args: any[]) => {
+            const initialFunction: any = build[fn as keyof PluginBuild];
```
Ayc0, line 60: same here
```suggestion
            const initialFunction: any = build[fn];
```

packages/esbuild-plugin/package.json
```diff
@@ -0,0 +1,42 @@
...
+    "name": "@datadog/esbuild-plugin",
+    "packageManager": "yarn@4.0.2",
+    "license": "MIT",
+    "author": "Datadog",
+    "main": "./src/index.ts",
+    "publishConfig": {
+        "access": "public"
+    },
+    "exports": {
+        ".": {
```
Ayc0, line 11: Same in this package.json (you don't need the import / require)

packages/tools/src/commands/oss/index.js
```diff
@@ -0,0 +1,221 @@
...
+
+const templates = require('./templates');
+
+const execFileP = promisify(execFile);
+const maxBuffer = 1024 * 1024;
+
+const execute = (cmd, args, cwd) => execFileP(cmd, args, { maxBuffer, cwd, encoding: 'utf-8' });
+
+const NAME = 'build-plugin';
+const ROOT = path.join(__dirname, '../../../../../');
```
Ayc0, line 24: You should be able to use `process.env.PROJECT_CWD` with yarn (no need to change this path every time the file moves around)

packages/webpack-plugin/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "packageManager": "yarn@4.0.2",
+    "private": true,
+    "license": "MIT",
+    "author": "Datadog",
+    "main": "./dist/packages/webpack-plugin/src/index.js",
+    "publishConfig": {
+        "access": "public"
+    },
+    "exports": {
+        ".": {
```
Ayc0, line 12: same here

Ayc0 (commented):
LGTM

One thing I just realized: in your `exports`, you referenced the `.ts` files. This will work in dev, but I don't think it'll work in prod when the package will be released as those files won't exist anymore

And this can break libraries / codebases using this library

packages/tools/src/commands/oss/index.js
yoannmoinet: 🤯

packages/webpack-plugin/src/modules.ts
```diff
@@ -0,0 +1,115 @@
...
+                this.storedDependents[dep] = this.storedDependents[dep] || new Set();
+                if (!this.storedDependents[dep].has(moduleName)) {
+                    this.storedDependents[dep].add(moduleName);
+                }
+            }
+        }
+
+        // Re-assign dependents to modules.
+        for (const storedDepName in this.storedDependents) {
+            if (Object.prototype.hasOwnProperty.call(this.storedDependents, storedDepName)) {
```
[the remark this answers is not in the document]
yoannmoinet: 👴

packages/core/src/hooks/datadog/metrics/esbuild.ts
yoannmoinet: Types say that it can't be `null`, but this is still best to have this safe guard.

packages/webpack-plugin/src/modules.ts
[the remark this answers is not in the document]
yoannmoinet: Needed `"lib": ["es2022"]` in `tsconfig.json` but it's worth it.

packages/assets/package.json
Ayc0: package.json should always be accessible
And if you want to remove those tsconfig and stuff, you can use either `files` in the package.json to opt in some, or use `.npmignore` to remove some on publish, see https://docs.npmjs.com/cli/v10/configuring-npm/package-json#files

packages/core/README.md
yoannmoinet: I'll fill it out in a different PR where I'll go around all docs.

packages/core/package.json
```diff
@@ -6,11 +6,42 @@
...
+                "require": "./dist/packages/core/src/*.js"
+            }
+        }
+    },
+    "files": [
+        "dist"
+    ],
     "scripts": {
-        "typecheck": "tsc --noEmit"
+        "build": "yarn clean; tsc",
```
Ayc0, line 40: shouldn't this be `yarn clean && tsc`?

packages/esbuild-plugin/package.json
```diff
@@ -0,0 +1,52 @@
...
+        "esbuild": "0.20.2"
+    },
+    "peerDependencies": {
+        "esbuild": "*"
+    },
+    "peerDependenciesMeta": {
+        "esbuild": {
+            "optional": true
+        }
+    }
```
Ayc0, line 51: Is it still optional? (as this is the esbuild-plugin)

packages/webpack-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "webpack": "5.49.0"
+    },
+    "peerDependencies": {
+        "webpack": "*"
+    },
+    "peerDependenciesMeta": {
+        "webpack": {
+            "optional": true
+        }
+    }
```
Ayc0, line 52: Is it still optional as thi sis the webpack plugin?

packages/webpack-plugin/package.json
```diff
@@ -0,0 +1,53 @@
...
+    },
+    "devDependencies": {
+        "typescript": "5.4.3"
+    },
+    "dependencies": {
+        "@datadog/build-plugins-core": "workspace:*",
+        "webpack": "5.49.0"
+    },
+    "peerDependencies": {
+        "webpack": "*"
```
Ayc0, line 46: nit: maybe we want to say 5.x.x as Idk if this works with webpack 4 (and def. doesn't work with webpack 3)

Ayc0 (approved):

packages/core/package.json
yoannmoinet: Seems to give the same result.

packages/webpack-plugin/package.json
yoannmoinet: Good point, it should support webpack 4 and 5.
I'll update.

packages/webpack-plugin/package.json
yoannmoinet: Good point, will update.

packages/esbuild-plugin/package.json
yoannmoinet: Good point, will update.

PR #64, opened 2024-04-26
[v2.0.0] Use unplugin

yoannmoinet:


.eslintrc.js
```diff
@@ -149,12 +149,7 @@ module.exports = {
         'no-await-in-loop': 'warn',
         'no-compare-neg-zero': 'error',
         'no-cond-assign': ['error', 'always'],
-        'no-console': [
-            'error',
-            {
-                allow: ['warn', 'error', 'info'],
-            },
-        ],
+        'no-console': 0,
```
Ayc0, line 152: prefer `'off'` (I can never remember is `0` = `off` or `error`)

Ayc0: Also if it's off, maybe we can just remove the line

packages/factory/package.json
```diff
@@ -0,0 +1,15 @@
+{
+    "name": "@dd/factory",
```
Ayc0, line 2: Do we have the rights on `@dd`?

packages/plugins/telemetry/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "author": "Datadog",
+    "exports": {
+        ".": "./src/index.ts",
+        "./esbuild-plugin/*": "./src/esbuild-plugin/*.ts",
+        "./webpack-plugin/*": "./src/webpack-plugin/*.ts",
+        "./common/*": "./src/common/*.ts",
+        "./common/metrics/*": "./src/common/metrics/*.ts",
+        "./common/output/*": "./src/common/output/*.ts",
+        "./*": "./src/*.ts"
+    },
```
Ayc0, line 15: You'll need a `publishConfig` otherwise those won't be valid for the built code

Also isn't `"./**": "./src/**.ts"` already working for all those options? (except `.`)

And you may want to add `./package.json` (common to include)

.eslintrc.js
yoannmoinet: Did it in https://github.com/DataDog/build-plugin/pull/65

packages/factory/package.json
yoannmoinet: This is not published, only internal dependencies.

packages/plugins/telemetry/package.json
yoannmoinet: Most of the publication work is done in https://github.com/DataDog/build-plugin/pull/65

Ayc0 (approved):
As most of my comments are solved in #65, let's merge this for now

PR #65, opened 2024-04-30
[v2.0.0] Package the plugins

yoannmoinet:
### What and why?

We want to publish 2 packages and have them work as expected OOTB.

- `@datadog/webpack-plugin`
- `@datadog/esbuild-plugin`

### How?

We now bundle the published packages with rollup, using a shared configuration.
We had to refactor what was published and what not.
We've assessed the usage from web-ui using prerelease versions.

Fixes https://github.com/DataDog/build-plugin/issues/59

Tested it in https://github.com/DataDog/web-ui/pull/136308


yoannmoinet:
Hello World

.eslintrc.js
```diff
@@ -376,5 +375,16 @@ module.exports = {
...
         },
+        {
+            files: [
+                'rollup.config.mjs',
+                'packages/core/**/*',
+                'packages/webpack-plugin/**/*',
+                'packages/esbuild-plugin/**/*',
+            ],
+            rules: {
+                'import/no-extraneous-dependencies': 0,
```
Ayc0, line 386:
```suggestion
                'import/no-extraneous-dependencies': 'error',
```

packages/core/src/types.ts
```diff
@@ -244,7 +244,6 @@ export type GetPlugins<T> = (options: T) => UnpluginOptions[];
 export interface GetPluginsOptions {
     auth: {
         apiKey: string;
-        appKey: string;
```
Ayc0, line 247: Why did you remove the `appKey`?

packages/esbuild-plugin/package.json
```diff
@@ -1,40 +1,71 @@
 {
     "name": "@datadog/esbuild-plugin",
     "packageManager": "yarn@4.0.2",
+    "version": "0.0.13-4",
     "license": "MIT",
     "author": "Datadog",
+    "description": "Datadog ESBuild Plugin.",
```
Ayc0, line 7: Nit: you may want to drop the `.` at the end of all your descriptions

packages/esbuild-plugin/package.json
```diff
@@ -1,40 +1,71 @@
...
+    "repository": {
+        "type": "git",
+        "url": "https://github.com/DataDog/build-plugin",
+        "directory": "packages/esbuild-plugin"
+    },
     "main": "./dist/src/index.js",
+    "module": "./dist/src/index.mjs",
     "exports": {
         ".": "./src/index.ts",
         "./*": "./src/*.ts"
```
Ayc0, line 25: Shouldn't we drop the `./*` (as it's not overridden in the `publishConfig` which means that it won't work for built code)

packages/core/src/types.ts
yoannmoinet: It was not used anywhere.

packages/esbuild-plugin/package.json
```diff
@@ -1,40 +1,71 @@
...
+                "types": "./dist/src/index.d.ts"
+            }
         }
     },
     "files": [
         "dist"
     ],
     "scripts": {
-        "build": "yarn clean && tsc",
+        "build": "yarn clean && rollup --config rollup.config.mjs",
```
Ayc0, line 43: If the code is built by `rollup`, you way want to add `"moduleResolution": "bundler"` & `"module": "NodeNext"` in your TSConfig.json

packages/esbuild-plugin/package.json
```diff
@@ -1,40 +1,71 @@
...
         "@dd/factory": "workspace:*",
-        "esbuild": "0.20.2"
+        "@dd/tools": "workspace:*",
+        "@rollup/plugin-babel": "6.0.4",
+        "@rollup/plugin-commonjs": "25.0.7",
+        "@rollup/plugin-json": "6.1.0",
+        "@rollup/plugin-node-resolve": "15.2.3",
+        "@types/babel__core": "^7",
+        "@types/babel__preset-env": "^7",
+        "esbuild": "0.20.2",
```
Ayc0, line 61: Note: you could also use `esbuild` as your main builder instead of having to install:
- rollup
- all the rollup plugins
- babel
- all the babel plugins

You can follow this guide for how to setup esbuild: https://esbuild.github.io/getting-started/#bundling-for-node

packages/plugins/telemetry/package.json
```diff
@@ -28,11 +35,7 @@
         "esbuild": "0.20.2"
     },
     "peerDependencies": {
+        "esbuild": "*",
         "webpack": "*"
```
Ayc0, line 39: You should add min versions, like `"esbuild": ">=1.2.3"`

packages/tests/package.json
```diff
@@ -14,16 +21,17 @@
         "test": "NODE_OPTIONS=\"--openssl-legacy-provider ${NODE_OPTIONS:-}\" jest --verbose"
     },
     "dependencies": {
-        "@datadog/build-plugins-core": "workspace:*",
-        "@datadog/esbuild-plugin": "workspace:*",
-        "@datadog/webpack-plugin": "workspace:*",
+        "@datadog/esbuild-plugin": "workspace:0.0.13-4",
+        "@datadog/webpack-plugin": "workspace:0.0.13-4",
```
Ayc0, line 25: To note: you shouldn't used pinned version, but `workspace:*`

yarn will use the version automatically of the other package on publish 
And if you publish them in order, it should be okay (this is what I do with my packages and it works amazingly good)

packages/webpack-plugin/package.json
```diff
@@ -1,41 +1,71 @@
...
+        "rollup": "4.17.1",
+        "rollup-plugin-dts": "6.1.0",
+        "rollup-plugin-esbuild": "6.1.1",
+        "typescript": "5.4.3"
     },
     "peerDependencies": {
         "webpack": ">= 4.x < 6.x"
-    }
+    },
+    "stableVersion": "0.0.12-dev"
```
Ayc0, line 70: what is `stableVersion`?

Ayc0 (approved):

.eslintrc.js
yoannmoinet: Fixed in https://github.com/DataDog/build-plugin/pull/67
In this case, you wanted it to be `'off'` right?

packages/esbuild-plugin/package.json
yoannmoinet: Will do, but why?

packages/esbuild-plugin/package.json
yoannmoinet: Good catch.

packages/esbuild-plugin/package.json
yoannmoinet: Used `"preserve"` as discussed, except for the tests package.

packages/esbuild-plugin/package.json
yoannmoinet: I'm fine with rollup right now, will change if we get to a point where it's necessary to change.

packages/plugins/telemetry/package.json
yoannmoinet: Good call.

packages/esbuild-plugin/package.json
Ayc0: Almost no-one uses `.` at the end of descriptions

packages/webpack-plugin/package.json
Ayc0: TIL

PR #67, opened 2024-05-21
[v2.0.0] Onboarding and documentation

yoannmoinet:
### What and why?

Update all the documentation and add more.
Also create some CLIs to help with the DX.


.eslintrc.js
```diff
@@ -373,6 +373,7 @@ module.exports = {
                 'global-require': 0,
                 'import/no-dynamic-require': 0,
                 'import/no-extraneous-dependencies': 0,
+                'no-await-in-loop': 0,
```
Ayc0, line 376:
```suggestion
                'no-await-in-loop': 'off',
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Install dependencies
+
+<details>
+<summary>Use Volta (recommended)</summary>
+
+```bash
+brew install volta
+volta setup
+volta install node yarn
```
Ayc0, line 17: You need to be in the right folder to run this

CONTRIBUTING.md
```diff
@@ -0,0 +1,150 @@
...
+
+</details>
+
+> [!NOTE]
+> Do not worry about the version, it's dealt with by the repo.
+
+Ensure dependencies are up to date in the repository.
+
+```bash
+cd build-plugins
```
Ayc0, line 37: Related to https://github.com/DataDog/build-plugin/pull/67/files#r1616167814

I think it's fair to assume that you are in the right folder. Maybe we can drop this `cd`

packages/tools/src/commands/create-plugin/index.ts
```diff
@@ -0,0 +1,120 @@
...
+import fs from 'fs-extra';
+import path from 'path';
+
+import { ROOT } from '../../constants';
+import { green, slugify, runAutoFixes } from '../../helpers';
+import type { Context, Plugin } from '../../types';
+import { updateFiles } from '../integrity/files';
+import { updateReadmes } from '../integrity/readme';
+
+import { getFiles } from './templates';
```
Ayc0, line 17: Avoid top level `import`s as you're using a require on each file for `yarn cli`

so all those imports are slowing down the other commands (even if you're not running them)

.github/workflows/ci.yaml
```diff
@@ -34,8 +34,8 @@ jobs:
       with:
         node-version: '18.19.0'
     - run: yarn install
-    - run: yarn oss
-    - run: git diff --exit-code && git diff --cached --exit-code || (echo "Please run 'yarn oss' and commit the result." && exit 1)
+    - run: yarn cli integrity
+    - run: git diff --exit-code && git diff --cached --exit-code || (echo "Please run 'yarn cli integrity' and commit the result." && exit 1)
```
Ayc0, line 38: Shouldn't we run both `oss` and `integrity`?

packages/tools/src/commands/integrity/index.ts
```diff
@@ -0,0 +1,61 @@
...
+                ...plugin,
+                slug: plugin.location.split('/').pop() as string,
+            }));
+    }
+
+    async execute() {
+        // Load all the plugins.
+        const plugins = await this.getPlugins();
+
+        const errors = [];
```
Ayc0, line 45: You may want to type this, otherwise it'll be `unknown[]` or `never[]` IIRC

packages/tools/package.json
```diff
@@ -15,19 +15,25 @@
         "./*": "./src/*"
     },
     "scripts": {
-        "cli": "node ./src/index.js"
+        "cli": "ts-node ./src/index.ts"
```
Ayc0, line 18: You can speed this up by running `ts-node -T` or `ts-node --transpileOnly`

Or even better `ts-node --swc`

packages/tools/src/commands/create-plugin/index.ts
yoannmoinet: Oh yeah I wanted to move them in a separate file, I'll do it.

.github/workflows/ci.yaml
yoannmoinet: `oss` is ran at the end of `integrity` part of the `runAutoFixes`.

packages/tools/package.json
yoannmoinet: NICE!

Ayc0 (approved):
Love to see those 🤩

PR #68, opened 2024-05-21
[v2.0.0] Issue fixes

yoannmoinet:
### What and why?

It unifies the logging across all the plugins.
And move some configuration around.

Fixes https://github.com/DataDog/build-plugin/issues/58


packages/core/src/log.ts
```diff
@@ -0,0 +1,38 @@
...
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import c from 'chalk';
+
+import type { LogLevel } from './types';
+
+const log = (text: string, level: LogLevel, type: LogLevel = 'debug', name?: string) => {
+    let color = c;
+    // eslint-disable-next-line no-console
```
Ayc0, line 11: You should disable this rule fully in this file with `// eslint-disable no-console` at the top and remove all the inline ones

packages/plugins/telemetry/src/common/output/text.ts
```diff
@@ -36,37 +37,43 @@ const sortDesc = (attr: any) => (a: any, b: any) => {
     }
 };
 
-const render = (values: any[], renderValue: (arg: any) => string) => {
+const getOutput = (values: any[], renderValue: (arg: any) => string): string => {
+    let st = '';
```
Ayc0, line 41: can you rename `st` to either `output` or `stdout`?

PR #69, opened 2024-05-30
[plugin] Add error tracking

yoannmoinet:
### What and why?

Adding the new `error-tracking` plugin.

### How?

Run `yarn cli create-plugin` and implement first feature, sourcemaps upload.


yoannmoinet:
Will do a different one.

PR #70, opened 2024-05-30
[integrity] add more integrity checks

yoannmoinet:
### What and why?

Ensure new plugins are following our standards.

### How?

We now generate a full configuration in the main README.
We also error if any plugin misses a title, description and configuration in their READMEs.
We now verify that every plugin is owned by someone.
Also renamed everything to `DataDog/build-plugins`.

### DIFF
--- a/packages/tools/src/commands/integrity/readme.ts
+++ b/packages/tools/src/commands/integrity/readme.ts
@@ -69,9 +69,14 @@ const getPluginMetadata = async (plugin: Workspace): Promise<PluginMetadata> => 
     // Load plugin's README.md file.
     const readmePath = path.resolve(ROOT, plugin.location, 'README.md');
     const readme = fs.readFileSync(readmePath, 'utf-8');
+
     // Get the title and the first paragraph.
+    // Catch the first title and remove the `MD_TOC_OMIT_KEY` from it.
     const title = readme.match(/# (.*)/)?.[1].replace(` ${MD_TOC_OMIT_KEY}`, '') || '';
+    // Catch the first line of text after the title.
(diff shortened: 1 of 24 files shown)


packages/tools/src/commands/integrity/readme.ts
```diff
@@ -51,19 +64,25 @@ const getReadmeToc = (readmeContent: string) => {
...
+const getPluginMetadata = async (plugin: Workspace): Promise<PluginMetadata> => {
     const { CONFIG_KEY } = await require(path.resolve(ROOT, plugin.location, 'src/constants.ts'));
     // Load plugin's README.md file.
     const readmePath = path.resolve(ROOT, plugin.location, 'README.md');
     const readme = fs.readFileSync(readmePath, 'utf-8');
     // Get the title and the first paragraph.
     const title = readme.match(/# (.*)/)?.[1].replace(` ${MD_TOC_OMIT_KEY}`, '') || '';
     const intro = readme.match(/# .*\n\n(.*)/)?.[1] || '';
-    return { title, intro, key: CONFIG_KEY };
+    const config = readme.match(/## Configuration[\s\S]*?```[^\n\r]+\n([\s\S]*?)\n```/)?.[1] || '';
```
elbywan, line 75: got a bit puzzled by this `[\s\S]*?` but If understood correctly the diff with `.*` is that it matches the line breaks right?

elbywan (approved):
LGTM 👍

packages/tools/src/commands/integrity/readme.ts
yoannmoinet: Yes, exactly, I'll add a comment to clarify this.

PR #76, opened 2024-06-03
[create-plugin] Better bootstrapping

yoannmoinet:
### What and why?

It's a bit confusing when creating a new plugin what to pick from the wizard.

### How?

Giving more choices and explanation over what each hook is doing.

image


Miz85 (approved):
👍🏼  LGTM! Nice addition to the build plugin creation flow!
Maybe we could also add a link to docs that describe each hook in more details if it exists!

yoannmoinet:
@Miz85 I added a link in the template it creates for each hook/bundler.

image

elbywan (approved):

PR #77, opened 2024-06-03
[tests] Move existing mocks

yoannmoinet:
### What and why?

Existing mocks and helpers are specific to our telemetry plugin.

### How?

Move the relevant files to their relevant place and update some docs.


packages/core/src/log.ts
```diff
@@ -28,11 +28,12 @@ const log = (text: string, level: LogLevel, type: LogLevel = 'debug', name?: str
...
     ) {
-        logFn(`${prefix}${color(text)}`);
+        const content = typeof text === 'string' ? text : JSON.stringify(text, null, 2);
+        logFn(`${prefix}${color(content)}`);
     }
 };
 
 export const getLogFn =
     (level: LogLevel = 'warn', name?: string) =>
-    (text: string, type: LogLevel = 'debug') =>
```
yoannmoinet, line 37: This is a fix for the log function, not super related with the tests move.

elbywan (approved):

PR #82, opened 2024-06-20
[core] core plugins update

yoannmoinet:
### What and why?

Still building up https://github.com/DataDog/build-plugins/pull/78 and trying to split it as much as possible to keep the review complexity down.

### How?

This one can be summed up as:

- Some small config updates that were missed in previous PRs.
- Refactor of the internal plugin mechanism and move some types.
- Implementation of internal git plugin that adds some git metadata to the build.
- Update some test helpers.


packages/core/src/types.ts
```diff
@@ -2,20 +2,56 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2019-Present Datadog, Inc.
 
-import type { UnpluginOptions } from 'unplugin';
+// #imports-injection-marker
```
elbywan, line 5: ⛏️ **small nit**: I don't think this would be easy for a new contributor to understand that the code between these 2 markers gets replaced when the `yarn cli integrity` tool runs so it might be nice to add a comment or something? 
(also I could not find in the contributing.md/readme.md a mention about when to run this tool but maybe it's not meant to?)

elbywan (approved):
LGTM 👍

PR #83, opened 2024-06-25
[RUM] Remove form data dependency

yoannmoinet:
### What and why?

Remove dependency on `form-data` as NodeJS is supposed to handle `FormData` and `fetch` natively.

### How?

I don't know yet.

Blocking error when running the `sendSourcemaps` test:

```
node:internal/deps/undici/undici:11730
    Error.captureStackTrace(err, this);
          ^

TypeError: Response body object should not be disturbed or locked
    at Object.fetch (node:internal/deps/undici/undici:11730:11)

Node.js v18.19.0
```


yoannmoinet:
Already added in #78

PR #84, opened 2024-07-08
[core] More global contexts

yoannmoinet:
Stacked on https://github.com/DataDog/build-plugins/pull/85

### What and why?

We need a way to communicate the list of files produced at the end of the build.

### How?

We're using the global context plugin to gather the list from webpack and esbuild (since they are the only two bundles we support right now).
And add it to the global context.

### DIFF
--- a/packages/tests/src/helpers/configBundlers.ts
+++ b/packages/tests/src/helpers/configBundlers.ts
@@ -23,6 +23,7 @@ export const getWebpackOptions = (
 
     return {
         entry: defaultEntry,
+        mode: 'production',
         output: {
             path: path.join(defaultDestination, 'webpack'),
             filename: `[name].js`,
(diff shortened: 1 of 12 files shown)


elbywan (approved):

packages/tests/src/helpers/configBundlers.ts
```diff
@@ -45,7 +46,8 @@ export const getWebpack4Options = (
     const plugin = datadogWebpackPlugin(newPluginOptions) as unknown;
 
     return {
-        entry: defaultEntry,
+        // Somehow webpack4 doesn't find @dd/tests/fixtures/index.js
```
yoannmoinet, line 49: I think it's simply that it doesn't support pnp without a plugin.

PR #90, opened 2024-08-13
[telemetry] Rewrite Telemetry module to be more universal

yoannmoinet:
### What and why?

The telemetry module is difficult to work with, hard to implement in other bundlers and not super configurable.
We need to widen the commun trunk of generated metrics and open its configuration to enable more granular processes.

### How?

- Create a global, universal report for builds in our global context plugin.
- Update the telemetry module to be universal for the most part (everything except plugins' tracing)


yoannmoinet:
Will re-open in a separate stacked PR.

PR #96, opened 2024-09-09
Custom Plugins

yoannmoinet:
### What and why?

Allowing the user to inject their own plugins in their configuration in order to leverage the shared context of our plugin.

### How?

Add a new `customConfig` configuration.
More details in the [README](https://github.com/DataDog/build-plugins/blob/a95eea3a641bc751a934a26d680da5d55663a549/README.md#customplugins).

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,7 @@ datadogWebpackPlugin({
     auth?: {
         apiKey?: string;
     };
+    customPlugins?: (options: Options, context: GlobalContext) => UnpluginPlugin[];
     logLevel?: 'debug' | 'info' | 'warn' | 'error' | 'none';
     rum?: {
         disabled?: boolean;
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -251,6 +252,7 @@ datadogWebpackPlugin({
     auth?: {
         apiKey?: string;
     };
+    customPlugins?: (options: Options, context: GlobalContext) => UnpluginPlugin[];
```
elbywan, line 255: I don't think that this type is correct, `customPlugins` expects an array of functions

README.md
```diff
@@ -296,18 +298,86 @@ In order to interact with Datadog, you have to use [your own API Key](https://ap
...
+
+This is a way for you to inject any [Unplugin Plugin](https://unplugin.unjs.io/guide/) you want.
+
+It's particularly useful  to use our global, shared context of the main plugin.
+
+```typescript
+{
+    customPlugins: [
+        (options, context) => {
+            return {
```
elbywan, line 313: I think that the return type is supposed to be an array of plugins

packages/core/src/types.ts
```diff
@@ -102,6 +103,7 @@ export interface Options extends GetPluginsOptions {
     [rum.CONFIG_KEY]?: RumOptions;
     [telemetry.CONFIG_KEY]?: TelemetryOptions;
     // #types-injection-marker
+    customPlugins?: GetCustomPlugins<Options>[];
```
elbywan, line 106: I don't know if accepting an array of functions with each of them returning an array of plugins is necessary 🤔 
IMO having a single function returning an array of plugins should be enough, is there a particular reason to not do so?

```diff
// This is what I mean:
- customPlugins?: GetCustomPlugins<Options>[];
+ customPlugins?: GetCustomPlugins<Options>
```

elbywan (commented):
Left a couple of comments 💬

README.md
yoannmoinet: Oh right, I missed a `[]`.
Good catch !

packages/core/src/types.ts
yoannmoinet: I thought about it too, but I didn't go into it deeply enough.
So no strong opinion on it at all, and I agree with you that a single function returning an array would be plenty enough.

I'll updated.

elbywan (approved):

README.md
```diff
@@ -296,18 +298,86 @@ In order to interact with Datadog, you have to use [your own API Key](https://ap
...
+    start: number;
+    // Available from the initialization.
+    version: string;
+}
+```
+
+Your function will need to return an array of [Unplugin Plugins definitions](https://unplugin.unjs.io/guide/#supported-hooks).
+
+> [!NOTE]
+> Some parts of the context are only available after certain hooks.
```
egeriis, line 373: Is this something that can be elaborated upon?

yoannmoinet: For each entry there is a comment "Available from *" to give a better idea of which hook they're available from.
I'm not sure what I could add to make it clearer, any suggestion?

PR #101, opened 2024-09-30
[Doc] Add some doc

yoannmoinet:
### What and why?

We were missing some documentation.

### How?

- Add documentation about testing.
- Add documentation about internal plugins.
- Fix generation of ToC.
- Factorise Global Context types.
- Some small updates


README.md
```diff
@@ -324,57 +324,84 @@ Your function will receive two arguments:
...
     // More details on the currently running bundler.
     bundler: {
         name: string;
         fullName: string; // Including its variant.
-        outDir: string;
+        outDir: string; // Output directory
+        // Added in `buildStart`.
         rawConfig?: any;
-        variant?: string; // Major version of the bundler (webpack 4, webpack 5)
+        variant: string; // Major version of the bundler (webpack 4, webpack 5)
```
elbywan, line 341: might be nice to tell the user that this will be an empty string for other bundlers (IIUC)

```ts
const variant =
        meta.framework === 'webpack' ? (meta.webpack.compiler['webpack'] ? '5' : '4') : '';
```

elbywan (approved):

README.md
yoannmoinet: Good call.

PR #103, opened 2024-10-03
[CLI] Fix create plugin wizard

yoannmoinet:
### What and why?

`yarn cli create-plugin` drifted a bit given we're not using it that much.

We need to update/fix it, have it follow current standards and test it to ensure it remains as up to date as possible with the reste of the codebase's growth.

### How?

- Fix bugs.
- Update contents.
- Test it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,7 +338,7 @@ type GlobalContext = {
         outDir: string; // Output directory
         // Added in `buildStart`.
         rawConfig?: any;
-        variant: string; // Major version of the bundler (webpack 4, webpack 5)
+        variant: string; // Major version of the bundler (webpack 4, webpack 5), empty string otherwise.
     };
     // Added in `writeBundle`.
     build: {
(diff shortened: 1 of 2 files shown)


packages/tools/tsconfig.json
```diff
@@ -5,8 +5,11 @@
         "esModuleInterop": true,
         "module": "CommonJS",
         "moduleResolution": "Node",
-        "rootDir": "./",
-        "outDir": "./dist"
+        "outDir": "./dist",
+        "paths": {
+            "@dd/tools/*": ["./src/*"],
+            "@dd/core/*": ["../core/src/*"]
+        }
```
yoannmoinet, line 12: This config is slightly different from the other workspaces because we use `ts-node` to run the CLI.

I'd love to have the exact same config everywhere, but it bite me a few times already trying to unify this. 
Taking me hours every time I try to find a way, always reverting everything in the end.

Open to discuss if you have a very clear vision of it.

elbywan (approved):

PR #106, opened 2024-10-25
[bundling] Change bundling

yoannmoinet:
### What and why?

We should not bundle external dependencies in our published packages.
We still need to bundle though, because we have shared internal dependencies cross workspaces.

Also update our dependencies.

### How?

1. Dedupe dependencies.
2. Upgrade `rollup` (+ plugins), `vite`, `unplugin` and `esbuild`.
3. Remove `fs-extra` and replace it with helpers and native `fs`.
4. Update Jest's setup to avoid collision with mocks (of `fs` in `@dd/core/helpers` specifically).
    a. Move the setup files in a dedicated folder, and apply an eslint rule on it to avoid any internal imports in it.
5. Fix collisions of the injected file in `esbuild`.
6. Explicitly list dependencies of the packages we publish (based on what they internally depend on).
7. Add a new step in `yarn cli integrity` to verify that we have all the necessary dependencies in the published package.
8. Add `--cleanup=0` and `--build=1` flags for running tests.
    a. Jest actually does not support custom empty parameters and threw when I tried to add `--build`. It didn't throw with `--no-cleanup` because it's using `yargs` that automatically changed it to `--cleanup=0` behind the scene, which passed Jest's validation of accepting custom params with values. So I changed it to `--cleanup=0` for consistency with the new `--build=1`.
    b. `--build=1` will build all the bundlers' packages necessary prior to the test run.

> [!WARNING]
> - [x] Waiting after https://github.com/unjs/unplugin/pull/431 to be merged first


packages/tests/src/helpers/constants.ts
```diff
@@ -20,7 +20,12 @@ export const BUNDLER_VERSIONS: Record<BundlerFullName, string> = {
     webpack5: require('webpack5').version,
 };
 
-export const NO_CLEANUP = process.argv.includes('--no-cleanup');
+export const NO_CLEANUP = process.argv.includes('--cleanup=0');
```
yoannmoinet, line 23: Apparently Jest doesn't allow custom flags without values.
`--no-cleanup` was automatically transformed into `--cleanup=0` by Jest's dependency `yargs` so it passed through the validation because it has a value in the end.

So I transformed it to `--cleanup=0` to have continuity with `--build=1` because simply `--build` wasn't possible because Jest falgged it 😢.

packages/plugins/rum/src/sourcemaps/sender.ts
```diff
@@ -96,7 +96,9 @@ export const upload = async (
         return { errors, warnings };
     }
 
-    const queue = new PQueue({ concurrency: options.maxConcurrency });
+    // @ts-expect-error PQueue's default isn't typed.
+    const Queue = PQueue.default ? PQueue.default : PQueue;
+    const queue = new Queue({ concurrency: options.maxConcurrency });
```
yoannmoinet, line 101: This is the usual mess of highly opinionated Sindre Sorhus' packages.

packages/tests/src/helpers/generateMassiveProject.ts
```diff
@@ -71,8 +72,8 @@ export const generateProject = async (
     const target = path.join(__dirname, `../${PROJECT_PATH}`);
 
     // Clean previous project.
-    await fs.remove(target);
-    await fs.mkdir(target);
+    await rm(target);
+    await fs.mkdir(target, { recursive: true });
```
Ayc0, line 76: Nit: shouldn't you make a helper for this too?

yoannmoinet: That's fair.

Ayc0 (approved):

packages/tests/src/helpers/generateMassiveProject.ts
Ayc0: (no need to be done in this PR)

packages/plugins/rum/src/sourcemaps/sender.ts
elbywan: bummer :(

```
exports.default = PQueue;
```

packages/tools/src/commands/integrity/dependencies.ts
```diff
@@ -0,0 +1,190 @@
...
+                // All good.
+                outputLog += dim(`\n    "${depName}": "${depVersion}"`);
+            }
+        }
+        outputLog += '\n}';
+
+        if (!dependenciesMatch) {
+            // Log the error.
+            console.log(
+                `    Missmatch ${red('dependencies')} for ${red(bundler.name)}:\n${outputLog}`,
```
elbywan, line 180:
```suggestion
                `    Mismatch ${red('dependencies')} for ${red(bundler.name)}:\n${outputLog}`,
```

elbywan (approved):

PR #108, opened 2024-11-12
[refactor] Better integrate new internal plugins

yoannmoinet:
### What and why?

Now that we consider internal plugins almost the same as external plugins,
we need to better integrate how we create and verify them.

### How?

- Integrate their creation process in `yarn cli create-plugin`.
- Integrate their verification in `yarn cli integrity`.
- Update/simplify types.

> [!NOTE]
> I will remove `Check integrity of our files` from the expected checks once this PR is merged.


packages/tools/src/commands/create-plugin/hooks.ts
```diff
@@ -2,12 +2,45 @@
...
+        descriptions: [
+            'Create a customer facing plugin.',
+            'A single implementation for all bundlers.',
+            'It will use hooks that are supported by all bundlers.',
+        ],
+    },
+    bundler: {
+        name: `[${yellow('Discouraged')}] Bundler Specific Plugin`,
+        descriptions: [
+            'Create a plugin that will be customer facing.',
```
elbywan, line 29: **Super nit:** it feels weird to me to have different wording for the same thing ('Create a customer facing plugin.' / 'Create a plugin that will be customer facing.')

packages/tools/src/commands/integrity/files.ts
```diff
@@ -172,7 +244,7 @@ const verifyCodeowners = (plugins: Workspace[]) => {
 
 export const updateFiles = async (plugins: Workspace[]) => {
     const errors: string[] = [];
-    await updateFactory(plugins);
+    errors.push(...(await updateFactory(plugins)));
```
elbywan, line 247: Is there some reason that prevents doing this as a one liner?

```ts
const errors: string[] = await updateFactory(plugins);
```

packages/tools/src/commands/integrity/files.ts
```diff
@@ -77,32 +100,76 @@ const updateFactory = async (plugins: Workspace[]) => {
...
-        `;
 
-        // Only add helpers if they export them.
-        if (pluginExports.helpers && Object.keys(pluginExports.helpers).length) {
-            helperContent += `[${configKeyVar}]: ${camelCase}.helpers,`;
+            // Get the list of parameters to use when calling the internal plugin.
+            const paramsString = pluginExports[getFunction]
+                .toString()
+                // Find the list of arguments.
+                .match(/^\(([^)]+)\)/)[1]
```
elbywan, line 114: Is `pluginExports[getFunction]`always well formatted? If the argument list contains destructuring patterns or extra spacing / new lines it could mess up the regex matching and the subsequent split.

elbywan (approved):
🔥

packages/tools/src/commands/create-plugin/hooks.ts
yoannmoinet: Agreed 100%, I'll update.

packages/tools/src/commands/integrity/files.ts
yoannmoinet: Absolutely no reason at all, apart from some refactoring residue.
I'll update.

PR #109, opened 2024-11-14
[refactor] Better integrated logger.

yoannmoinet:
### What and why?

Our logger is a bit simplistic, it needs to be better integrated into the ecosystem.
Especially with the build-report to report errors and warnings.
As well as a simpler initialisation for plugins.

### How?

- Move the logger factory into the global context.
- Store all logs in build-report.
- Fix errors/warnings in esbuild and webpack build reports.
- Add tests.
- Factorise some mocks.
- Add some logs in hope to better understand test's race condition.


packages/core/src/types.ts
```diff
@@ -73,12 +74,21 @@ export type BundlerReport = {
 
 export type ToInjectItem = { type: 'file' | 'code'; value: string; fallback?: ToInjectItem };
 
+export type Logger = {
+    (text: any, type?: LogLevel): void;
```
Ayc0, line 78: why do you need this type?

packages/factory/src/helpers.ts
```diff
@@ -3,32 +3,100 @@
...
+
+            // Only log if the log level is high enough.
+            if (logPriority[type] >= logPriority[logLevel]) {
+                logFn(`${color(prefix)} ${content}`);
+            }
+        };
+
+        const logFn = (text: any, type: LogLevel = 'debug') => {
+            log(text, type);
+        };
```
Ayc0, line 67: I personally find it a bit weird to have both a top level fn and `logger.<level>()` fns

packages/plugins/build-report/src/esbuild.ts
```diff
@@ -68,8 +67,8 @@ export const getEsbuildPlugin = (context: GlobalContext, log: Logger): PluginOpt
             const entryNames = getEntryNames(entrypoints, context);
 
             build.onEnd((result) => {
-                context.build.errors = result.errors.map((err) => err.text);
-                context.build.warnings = result.warnings.map((err) => err.text);
+                context.build.errors.push(...result.errors.map((err) => err.text));
+                context.build.warnings.push(...result.warnings.map((err) => err.text));
```
Ayc0, line 71: Note: you're generating new arrays to immediately explode them
Idk if this can't be an issue for later arrays. If it is, it may be better to use forEach + push (but if it's not lgtm)

Ayc0 (approved):

packages/core/src/types.ts
yoannmoinet: It was to cover legacy usage.
But agreed (with your other comment as well), while I'm in there,
I'll remove this usage entirely.

PR #111, opened 2024-11-21
[bundler] Rspack

yoannmoinet:
### What and why?

We need to cover Rspack as we can't use webpack's plugin as-is, [webpack 5's API isn't compatible with Rspack's in a few areas we're using here and there](https://rspack.dev/api/plugin-api/index).

For example:
- Some hooks we're using.
- ModuleGraph isn't available.
- ...

### How?

- Add everything Rspack (logo, packages, helpers, ...).
- Publish a plugin for Rspack on `@datadog/rspack-plugin`.
- Unify webpack 4, 5 and Rspack implementations where possible, called it `xpack`.
  - I re-implemented the dependency graph of the build-report for rspack, and it seems to work the same for webpack 4 and 5, so I removed the webpack's one.

---

- [ ] I will test it on web-ui and report.

> [!NOTE]
> You need to run `yarn` to install binaries.


packages/tools/src/rollupConfig.mjs
```diff
@@ -19,11 +19,8 @@ export const bundle = (packageJson, config) => ({
     ...config,
     input: 'src/index.ts',
     external: [
-        // These are peer dependencies
-        'webpack',
-        'esbuild',
-        'vite',
-        'rollup',
+        // All peer dependencies are external dependencies.
+        ...Object.keys(packageJson.peerDependencies),
```
Ayc0, line 23: That is nice!

Ayc0 (approved):
Let's go! xpack I choose you!

image

README.md
```diff
@@ -1,6 +1,13 @@
 # Datadog Build Plugins <!-- #omit in toc -->
 
-A set of bundler plugins for <img src="packages/assets/src/webpack.svg" alt="Webpack" width="17" /> Webpack, <img src="packages/assets/src/vite.svg" alt="Vite" width="17" /> Vite, <img src="packages/assets/src/esbuild.svg" alt="ESBuild" width="17" /> ESBuild and <img src="packages/assets/src/rollup.svg" alt="Rollup" width="17" /> Rollup, to interact with Datadog directly from your builds.
+A set of bundler plugins for :
```
Ayc0, line 3: Nit: in english there is no space before :

```suggestion
A set of bundler plugins for:
```

README.md
```diff
@@ -1,6 +1,13 @@
...
+ - <img src="packages/assets/src/webpack.svg" alt="Webpack" width="17" /> Webpack
+ - <img src="packages/assets/src/vite.svg" alt="Vite" width="17" /> Vite
+ - <img src="packages/assets/src/esbuild.svg" alt="ESBuild" width="17" /> ESBuild
```
Ayc0, line 6: Nit: `esbuild` seems to always be lowercased
```suggestion
 - <img src="packages/assets/src/esbuild.svg" alt="ESBuild" width="17" /> esbuild
```

PR #112, opened 2024-11-26
[perf] Use Set instead of arrays for metrics aggregation

yoannmoinet:
### What and why?

This is slow and even crashes on very large lists.

### How?

Optimise the process using more `Set`s.

### DIFF
--- a/packages/plugins/telemetry/src/common/metrics/common.ts
+++ b/packages/plugins/telemetry/src/common/metrics/common.ts
@@ -4,10 +4,8 @@
 
 import type { TimingsMap, Metric } from '@dd/telemetry-plugin/types';
 
-export const getPlugins = (plugins: TimingsMap): Metric[] => {
-    const metrics: Metric[] = [];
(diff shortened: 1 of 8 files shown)


packages/plugins/telemetry/src/common/metrics/common.ts
```diff
@@ -26,67 +24,58 @@ export const getPlugins = (plugins: TimingsMap): Metric[] => {
                 hookDuration += duration;
                 pluginDuration += duration;
             }
-            metrics.push(
-                {
+            metrics
+                .add({
```
Ayc0, line 28: TIL set#add can be chained!

Ayc0 (approved):

packages/plugins/telemetry/src/common/metrics/common.ts
yoannmoinet: ME TOO, I didn't know that either.

PR #114, opened 2024-11-27
[fix] Small fixes

yoannmoinet:
### What and why?

https://github.com/DataDog/build-plugins/pull/112 introduced a bug in the sender of our telemetry plugin.

### How?

- Transform set into array right before sending metrics to DD.
- Add some comments here and there.
- Narrow down the entry filepath detection on xspack's build-report.

### DIFF
--- a/packages/plugins/build-report/src/helpers.ts
+++ b/packages/plugins/build-report/src/helpers.ts
@@ -22,6 +22,7 @@ import path from 'path';
 const EXTENSION_RX = /\.(?!.*(?:\.|\/|\\))(\w{1,})/g;
 
 // Will match any type of query characters.
+// ? or %3F or |.
 const QUERY_RX = /(\?|%3F|\|)+/gi;
 
 const getExtension = (filepath: string) => {
@@ -226,6 +227,7 @@ export const cleanReport = <T = string>(
 };
 
 // Clean a path from its query parameters and leading invisible characters.
+// Careful with this and webpack/rspack as loaders may add "|" before and after the filepath.
 export const cleanPath = (filepath: string) => {
     return (
         filepath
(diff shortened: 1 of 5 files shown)


packages/plugins/build-report/src/xpack.ts
```diff
@@ -367,7 +367,9 @@ export const getXpackPlugin =
                     // Filter the ones that includes the entry name.
                     .filter(
                         (f) => f.includes(name) || (entrypoint.name && f.includes(entrypoint.name)),
-                    )[0];
+                    )
+                    // Only keep JS files.
+                    .filter((f) => getType(f) === 'js')[0];
```
Ayc0, line 372: To note: if you use `filter` to then return the 1st element, you may want to use `.find` instead

packages/plugins/build-report/src/helpers.ts
```diff
@@ -22,6 +22,7 @@ import path from 'path';
 const EXTENSION_RX = /\.(?!.*(?:\.|\/|\\))(\w{1,})/g;
 
 // Will match any type of query characters.
+// ? or %3F or |.
```
Ayc0, line 25: what is `%3F`?

Also you should remove the `.` from this comment as I first understood that as "will match `|.`"

Ayc0 (approved):

packages/plugins/build-report/src/xpack.ts
yoannmoinet: Fair

packages/plugins/build-report/src/helpers.ts
yoannmoinet: `%3F` is `?` url encoded.

packages/plugins/build-report/src/xpack.ts
Ayc0: This will avoid creating a new array

PR #118, opened 2024-11-29
[sourcemaps] Move sourcemaps

yoannmoinet:
### What and why?

Sourcemaps upload is closer to Error Tracking than RUM.

### How?

Move everything into the Error Tracking plugin.


packages/plugins/error-tracking/package.json
```diff
@@ -1,15 +1,15 @@
 {
-    "name": "@dd/rum-plugin",
```
elbywan, line 2: it could be nice [mark the old package as deprecated](https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions) (maybe you're planning to do that already)

elbywan (approved):
LGTM 👍

packages/plugins/error-tracking/package.json
yoannmoinet: these are not published 😉

elbywan: woops right 👍

PR #119, opened 2024-11-29
[rum] Inject browser sdk

yoannmoinet:


packages/plugins/rum/README.md
```diff
@@ -0,0 +1,263 @@
+# Rum Plugin <!-- #omit in toc -->
```
Ayc0, line 1:
```suggestion
# RUM Plugin <!-- #omit in toc -->
```

README.md
```diff
@@ -276,6 +309,55 @@ datadogWebpackPlugin({
...
+
+<details>
+
+<summary>Configuration</summary>
+
+```typescript
+datadogWebpackPlugin({
+    rum?: {
+        disabled?: boolean,
+        sdk?: {
```
Ayc0, line 326: Maybe we could use the RUM type directly here, no?

packages/plugins/rum/src/built/rum-browser-sdk.ts
```diff
@@ -0,0 +1,17 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+/* global globalThis */
+
+import { datadogRum } from '@datadog/browser-rum';
+
+// To please TypeScript.
```
Ayc0, line 9: 😂

packages/plugins/rum/src/built/rum-browser-sdk.ts
```diff
@@ -0,0 +1,17 @@
...
+const globalAny: any = globalThis;
+
+// Also them to the global DD_RUM object.
+globalAny.DD_RUM = globalAny.DD_RUM || {};
+globalAny.DD_RUM = {
+    ...globalAny.DD_RUM,
+    ...datadogRum,
```
Ayc0, line 16: I don't have any knowledge on whether `window.DD_RUM` and `datadogRum` share the exact same API. I'll let @Miz85 review this one

packages/plugins/rum/src/index.ts
```diff
@@ -0,0 +1,55 @@
...
+    // Verify configuration.
+    const options = validateOptions(opts, log);
+
+    // NOTE: These files are built from "@dd/tools/rollupConfig.mjs" and available in the distributed package.
+    if (options.sdk) {
+        // Inject the SDK from the CDN.
+        context.inject({
+            type: 'file',
+            // Using MIDDLE otherwise it's not executed in context.
+            position: InjectPosition.MIDDLE,
```
Ayc0, line 41: What does `MIDDLE` mean exactly?

To my knowledge, RUM needs to be the 1st script executed to be able to track all resources properly

packages/plugins/rum/src/validate.ts
```diff
@@ -0,0 +1,103 @@
...
+            startSessionReplayRecordingManually: false,
+            storeContextsAcrossPages: false,
+            telemetrySampleRate: 20,
+            traceSampleRate: 100,
+            trackingConsent: 'granted',
+            trackLongTasks: false,
+            trackResources: false,
+            trackUserInteractions: false,
+            trackViewsManually: false,
+        };
```
Ayc0, line 93: Same here: Idk if we should re-code the SDK defaults. Is it possible to get them from the SDK package itself? (I'd prefer to avoid duplicating code, specially if they decide to change their defaults)

README.md
yoannmoinet: That's a very good idea, I'll check if I can.

packages/plugins/rum/README.md
```diff
@@ -0,0 +1,263 @@
...
+## Table of content <!-- #omit in toc -->
+
+<!-- This is auto generated with yarn cli integrity -->
+
+<!-- #toc -->
+-   [Configuration](#configuration)
+-   [Browser SDK Injection](#browser-sdk-injection)
+    -   [rum.sdk.applicationId](#rumsdkapplicationid)
+    -   [rum.sdk.clientToken](#rumsdkclienttoken)
+    -   [rum.sdk.site](#rumsdksite)
```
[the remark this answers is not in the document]
yoannmoinet: Makes sense, I'm afraid of getting out of sync, but I do prefer having a single source of truth.

packages/plugins/rum/src/validate.ts
yoannmoinet: Fair, I'll look into it.

packages/plugins/rum/README.md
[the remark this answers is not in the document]
Ayc0: If you want only 1 source of truth, you'd need to add a dependency on `@datadog/browser-rum` instead of checking the global `DD_RUM` (right now you're only using a devDep)

packages/plugins/rum/src/built/rum-browser-sdk.ts
```diff
@@ -0,0 +1,17 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+/* global globalThis */
+
+import { datadogRum } from '@datadog/browser-rum';
```
Ayc0, line 7: Shouldn't you add this to your peerDeps too?

Because otherwise customers won't be able to import this as this is in our devDeps

Ayc0: Also you may want to mark this as optional peerDep (for people using `DD_RUM` and not the npm package)

and if it's an optional peerDep, the import could fail. So you need to do
```js
let datadogRum = {}
try {
 datadogRum = require('@datadog/browser-rum').datadogRum
} catch {}
```

packages/plugins/rum/src/built/rum-browser-sdk.ts
yoannmoinet: From [the documentation website](https://docs.datadoghq.com/real_user_monitoring/browser/setup/client?tab=rum):

image

I'd expect the two to be the same.

packages/plugins/rum/src/index.ts
yoannmoinet: From the [README](https://github.com/DataDog/build-plugins/tree/yoann/v2.5.0/packages/plugins/injection#injection-plugin-):

> `InjectPosition.MIDDLE`: Added at the beginning of the entry file, within the context of the bundle.

So it's the first piece of code in the context of the bundle.
It would be similar to a NPM installation and usage I believe.

yoannmoinet: To note the diff with `START`:

> `InjectPosition.START`: Added at the very beginning of the bundle, outside any closure.

Meaning it won't bundle anything with it (dependencies for instance). Basically `START` acts like the Banner concept in the bundling world.

packages/plugins/rum/src/built/rum-browser-sdk.ts
yoannmoinet: I'll add a comment, but this is bundled outside of the plugin.
Only used in case of injection.

In short, when I bundle the plugin, I also bundle this file on the side.
And if the customer wants to inject the SDK, I add the content of that bundled file in the `MIDDLE` injection point.

Ayc0: > Only used in case of injection.

But in case of injection, **they** need to be able to import `@datadog/browser-rum`.  So it needs to be in **their** dependencies (thus the peerDep)

Ayc0: > And if the customer wants to inject the SDK

This is also why I recommended an optional peerDep
Maybe we don't need the try / catch though, you're right

yoannmoinet: It's not an external dependency, `@datadog/browser-rum` is already bundled in the injected code.

Ayc0: AH, it's fully pre-bundled, I see. I may need to revise other comments too then 👌

packages/plugins/rum/src/validate.ts
Ayc0: Edit: as you're fully pinning a version of the browser SDK, hard coding values is not needed (as you can decide to update those when updating the underlying RUM version)

Idk if you want to manually check that the default values are the same, so it could still be good, but it's not required

Ayc0 (approved):
Nice!

packages/plugins/rum/src/built/rum-browser-sdk.ts
```diff
@@ -0,0 +1,17 @@
...
+
+// To please TypeScript.
+const globalAny: any = globalThis;
+
+// Also them to the global DD_RUM object.
+globalAny.DD_RUM = globalAny.DD_RUM || {};
+globalAny.DD_RUM = {
+    ...globalAny.DD_RUM,
```
Ayc0, line 15: Now that I better understand the fact that we are using a pre-bundled version of the SDK, does it make sense to try to inject here both `...DD_RUM, ...datadogRum`? (As those 2 could be using different versions)

Why not just doing `globalAny.DD_RUM = datadogRum`? (also IIRC, even the npm datadogRum should write in DD_RUM 🤔)

packages/plugins/rum/src/built/rum-browser-sdk.ts
Miz85: Yes they are the same.

But we should choose on strategy for the injection and roll with that, why do we want to support both the CDN and the npm install methods?

elbywan (approved):
lgtm

README.md
```diff
@@ -101,6 +102,38 @@ Follow the specific documentation for each bundler:
...
+            sessionReplaySampleRate?: number;
+            sessionSampleRate?: number;
+            silentMultipleInit?: boolean;
+            site?: string;
+            startSessionReplayRecordingManually?: boolean;
+            storeContextsAcrossPages?: boolean;
+            telemetrySampleRate?: number;
+            traceSampleRate?: number;
+            trackingConsent?: 'granted' | 'not_granted';
+            trackLongTasks?: boolean;
```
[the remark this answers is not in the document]
yoannmoinet: Do you know which ones? Because they all are still in [the doc](https://docs.datadoghq.com/real_user_monitoring/browser/setup/client/?tab=rum#configuration).

.github/workflows/ci.yaml
```diff
@@ -123,8 +123,6 @@ jobs:
       if: steps.cache-build.outputs.cache-hit != 'true'
       run: yarn build:all
 
-    - run: yarn typecheck:all
-
```
yoannmoinet, line 127: This is done part of the `yarn cli integrity` now.

PR #121, opened 2024-12-09
[internal] Add more injection points

yoannmoinet:
### What and why?

Here we are again, rewriting the injection plugin for the 18th time.
We'd need to inject code banner and footer, but also within the context of the bundle, being able to add new `import` instructions for instance.

Especially for [[rum] Inject browser sdk](https://github.com/DataDog/build-plugins/pull/119)

### How?

Some small update and fixes:
- Factorise some helpers (esbuild entry parsing, unique id generation).
- Fix some working directory issue with rollup and vite.
- Detect duplicate plugin names, otherwise the previous ones get overridden by the new ones.

Big refactor of the injection plugin:
- Split implementation of the injection plugin.
  - Each bundler is very much different, especially with the new needs of positional injection.
  - Re-implement some banner plugins closer to the bundler's API, to better support banner and footer (esbuild, webpack, rspack).
- Create new injection points, `START`, `MIDDLE` and `END`.
- Add more tests.

> [!NOTE]
> Tests are refactored in https://github.com/DataDog/build-plugins/pull/123


packages/core/package.json
```diff
@@ -24,12 +24,14 @@
     },
     "dependencies": {
         "async-retry": "1.3.3",
-        "chalk": "2.3.1"
+        "chalk": "2.3.1",
+        "glob": "11.0.0"
```
Ayc0, line 28: FYI node now supports OOTB glob (but it requires specific node version)

I'm not asking to change it, but just sharing knowledge hehehe

yoannmoinet: Oh that's super nice, I didn't know.
I'm planning on bumping node version at some point this quarter.

packages/core/src/helpers.ts
```diff
@@ -25,12 +27,72 @@ export const formatDuration = (duration: number) => {
...
 
-export const getResolvedPath = (filepath: string) => {
-    try {
-        return require.resolve(filepath);
-    } catch (e) {
-        return filepath;
+// https://esbuild.github.io/api/#glob-style-entry-points
+const getAllEntryFiles = (filepath: string): string[] => {
+    if (!filepath.includes('*')) {
+        return [filepath];
```
Ayc0, line 33: I don't think this is enough to know if this is a valid path / something that has to be globed. For instance `hello.{js,ts}` has to go through `glob`

packages/core/package.json
Ayc0: Edit: it's still experimental TIL – [link](https://nodejs.org/api/fs.html#fsglobpattern-options-callback)

image

packages/core/src/helpers.ts
```diff
@@ -25,12 +27,72 @@ export const formatDuration = (duration: number) => {
...
+    const resolutionErrors: string[] = [];
+
+    if (Array.isArray(entryPoints)) {
+        for (const entry of entryPoints) {
+            const fullPath = entry && typeof entry === 'object' ? entry.in : entry;
+            entryPaths.push({ path: fullPath });
+        }
+    } else if (typeof entryPoints === 'object') {
+        entryPaths.push(
+            ...Object.entries(entryPoints).map(([name, filepath]) => ({ name, path: filepath })),
```
Ayc0, line 58: Idk if this can happen, but if `entryPoints` is `null`, this will crash (typeof null is object)

packages/core/src/helpers.ts
```diff
@@ -25,12 +27,72 @@ export const formatDuration = (duration: number) => {
...
+        );
+    }
+
+    // Resolve all the paths.
+    const proms = entryPaths
+        .flatMap((entry) =>
+            getAllEntryFiles(entry.path).map<[{ name?: string; path: string }, string]>((p) => [
```
Ayc0, line 65: Pro tip: you can label tuples (not a request for change, just knowledge sharing):
`getAllEntryFiles(entry.path).map<[entry: { name?: string; path: string }, path: string]>((p) => [`

packages/plugins/bundler-report/src/index.ts
```diff
@@ -29,34 +45,106 @@ const xpackPlugin: (context: GlobalContext) => PluginOptions['webpack'] & Plugin
...
+        // The basic idea is to compare input vs output and keep the common part of the paths.
+        vite: {
+            config(config) {
+                if (config.root) {
+                    context.cwd = config.root;
+                } else {
+                    handleCwd(Array.from(directories), context);
+                }
+            },
+            ...rollupPlugin(),
```
Ayc0, line 145: You may want to put `...rollupPlugin()` before `config(config) {` to avoid having a field name `config` in `rollupPlugin()` overriding your config

packages/tests/src/_jest/helpers/mocks.ts
```diff
@@ -64,6 +65,46 @@ const logFn: Logger = {
...
+                pluginData: {},
+                path: getAbsolutePath(process.cwd(), filepath),
+            };
+        },
+        onStart: jest.fn(),
+        onEnd: jest.fn(),
+        onResolve: jest.fn(),
+        onLoad: jest.fn(),
+        onDispose: jest.fn(),
+        ...options,
```
Ayc0, line 87: You may want to move `...options` to the 1st element to avoid being able to override the other fields. And if you want that to be its behavior, you should add a comment saying "All the methods before can be overridden in `options`" or something like that

Ayc0 (approved):

packages/core/src/helpers.ts
yoannmoinet: Good catch, it can happen (according to the types).

packages/tests/src/_jest/helpers/mocks.ts
yoannmoinet: Changed `options` for `overrides`.

elbywan (approved):
🔥

PR #123, opened 2025-01-14
[tests] A better sandboxing of tests and builds

yoannmoinet:
### What and why?

After https://github.com/DataDog/build-plugins/pull/121 it proved to be super difficult not to leak in between builds.

We need to better sandbox our fixture projects and our builds from each others to avoid any leakage in between.

### How?

- Some inconsequential configuration updates.
- Better environment setup for Jest, where we can reliably instrument some flags
  - `--cleanup=0` to prevent deletion of the temporary directories where builds happen.
  - `--bundlers=webpack4,...` to only build for a specific bundler.
  - `--build` to force build the plugins before running the tests.
- Fixture projects are now isolated from the rest and are their own project (from yarn's perspective).
- Each time we trigger a build within a test (using `runBundlers()`) it will create a new temporary directory, copy the fixture projects in it, and run the build in isolation.


packages/tests/src/_jest/helpers/env.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (envLogs.length) {
+        console.log(`\n${envLogs.join('\n')}\n`);
+    }
+};
+
+export const getOutDir = (workingDir: string, folderName: string): string => {
+    return path.resolve(workingDir, `./dist/${folderName}`);
+};
+
+const FIXTURE_DIR = path.resolve(__dirname, '../fixtures');
```
Ayc0, line 111: Not a blocker, but `__dirname` doesn't exist in ESM, only in CJS. Idk if this can create issues or not

packages/tests/src/_jest/helpers/mocks.ts
```diff
@@ -315,4 +322,4 @@ export const filterOutParticularities = (input: File) =>
     // Exclude webpack buildin modules, which are webpack internal dependencies.
     !input.filepath.includes('webpack4/buildin') &&
     // Exclude webpack's fake entry point.
-    !input.filepath.includes('fixtures/project/empty.js');
+    !input.filepath.includes('fixtures/empty.js');
```
Ayc0, line 325: prefer `.endsWith` to `.includes`, to avoid something like `fixtures/empty.js/hello.js`

Ayc0 (approved):

packages/tests/src/_jest/helpers/env.ts
```diff
@@ -0,0 +1,135 @@
...
+        if (
+            !(REQUESTED_BUNDLERS as BundlerFullName[]).every((bundler) =>
+                FULL_NAME_BUNDLERS.includes(bundler),
+            )
+        ) {
+            throw new Error(
+                `Invalid "${red(`--bundlers ${REQUESTED_BUNDLERS.join(',')}`)}".\nValid bundlers are ${FULL_NAME_BUNDLERS.map(
+                    (b) => green(b),
+                )
+                    .sort()
```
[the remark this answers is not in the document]
yoannmoinet: absolutely not sure why I did this.

packages/tests/src/_jest/helpers/env.ts
yoannmoinet: I'll update once we transition to esm.

packages/tests/src/_jest/helpers/mocks.ts
yoannmoinet: fair.

yoannmoinet: 🤔 actually, given the context, I think we might also get `fixtures/empty.js?something` since it's within the bundling.
I've seen weird shit with the commonjs plugins for instance 😄
I prefer to filter out any mention of this path, which is also pretty much well under control.

elbywan (approved):

.github/workflows/ci.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
     - name: Install node
       uses: actions/setup-node@v4
       with:
-        node-version: '18.19.0'
+        node-version: '18.x'
```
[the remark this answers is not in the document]
yoannmoinet: super useful, thanks!

Issue #128, opened 2025-02-07
Vite build plugin: failed to load config


tywayne:
### Describe what happened
Starting an integration to a new app, ran in to issues during basic setup. 
Builds fail when using `@datadog/vite-plugin`, > v2.3.1. 
Build successfully in v2.3.0 or earlier.

### Steps to reproduce the issue:
- install `@datadog/vite-plugin` >2.3.1
- configure plugin in vite config
- run build

**Error message**
```txt
failed to load config from /apps/dashboard/vite.config.ts
error during build:
file://node_modules/@datadog/vite-plugin/dist/src/index.mjs:25
import vite from 'vite';
        ^^^^
 SyntaxError: The requested module 'vite' does not provide an export named 'default'
```

**Vite plugin config**
```ts
datadogVitePlugin({
    auth: {
        apiKey: *obscured*
    },
    rum: {
        sourcemaps: {
            dryRun: true,
            minifiedPathPrefix: '/assets/',
            releaseVersion:  *obscured*,
            service: 'dashboard',
        },
    },
}),
```

### Additional context

Node v18.20.6

PR #129, opened 2025-02-08
Update vite import

tywayne:
### What and why?

`vite` package does not include a default export, resulting in the error reported in #128 

### How?

Importing it as a namespace import using the `import * as vite` syntax


yoannmoinet:
Thank you for your PR and associated issue.
I'll have a look on Monday.

yoannmoinet:
I had to revert some commits on `master` to allow for this fix to pass before I release `v2.5.0` (visible in #130).

So I merged your commit directly in https://github.com/DataDog/build-plugins/commit/61613047fd874fae7558a211e793438159e7b622 and released it in [`v2.4.2`](https://github.com/DataDog/build-plugins/releases/tag/v2.4.2).

So sorry for the weird commit history, thanks a ton for your fix! 🙇

PR #133, opened 2025-02-17
[chore] Small updates

yoannmoinet:
### What and why?

Some small updates to lighten #134 and #137

### How?

- Change how we use the logger to simplify the documentation and reduce complexity of implementation. TLDR; Add `getLogger` on the `GlobalContext`.
- Move the `injections` state to its own plugin, for more simplicity and self contenance.
- Factorise some readmes types.


jakub-g (approved):

packages/factory/src/helpers.ts
```diff
@@ -119,8 +126,10 @@ export const getContext = ({
         build,
         // This will be updated in the bundler-report plugin once we have the configuration.
         cwd,
-        inject: (item: ToInjectItem) => {
-            injections.set(getUniqueId(), item);
+        getLogger: getLoggerFactory(build, options.logLevel),
+        // This will be updated in the injection plugin on initialization.
+        inject: () => {
+            throw new Error('Inject function called before it was initialized.');
```
Miz85, line 132: ❓**question** I'm not very familiar with the code but is that error caught and reported properly elsewhere or could it just break the build?

packages/factory/src/helpers.ts
```diff
@@ -25,11 +23,21 @@ const logPriority: Record<LogLevel, number> = {
     none: 4,
 };
 
-// Exported for testing.
+// Which separator to use for plugin names.
+export const NAME_SEP = '>';
```
Miz85, line 27: nit: 

```suggestion
export const PLUGINS_NAME_SEP = '>';
```

jakub-g (approved):

packages/factory/src/helpers.ts
```diff
@@ -25,11 +23,21 @@ const logPriority: Record<LogLevel, number> = {
...
+
+const cleanName = (name: string) => {
```
Miz85, line 29: nit:
```suggestion
const cleanPluginsName = (name: string) => {
```

Miz85 (approved):

PR #134, opened 2025-02-17
[logging] Add meta reporting

yoannmoinet:
### What and why?

We need a way to better understand how users are using the build plugins and with which bundler.

### How?

TLDR; Submit a log to Datadog with very basic informations, once the build has started.

image

[Logs are visible here.](https://app.datadoghq.com/logs?query=service%3Abuild-plugins&agg_m=count&agg_m_source=base&agg_t=count&cols=core_host%2Ccore_service%2Csource%2Cbundler.name&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1739617377098&to_ts=1739876577098&live=true)

### DIFF
--- a/packages/plugins/analytics/src/index.ts
+++ b/packages/plugins/analytics/src/index.ts
@@ -1,0 +1,69 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import { doRequest } from '@dd/core/helpers';
+import type { GlobalContext, PluginOptions } from '@dd/core/types';
+
+import { INTAKE_HOST, INTAKE_PATH, PLUGIN_NAME } from './constants';
+
+export { PLUGIN_NAME } from './constants';
+
+export const getAnalyticsPlugins = (context: GlobalContext): PluginOptions[] => {
+    const log = context.getLogger(PLUGIN_NAME);
+
+    context.sendLog = async (message: string, ctx: any = {}) => {
+        // Only send logs in production.
+        if (context.env !== 'production') {
+            return;
+        }
+
+        try {
+            const bundler = {
+                name: context.bundler.name,
+                version: context.bundler.version,
+            };
+
+            await doRequest({
+                // Don't delay the build too much on error.
+                retries: 2,
+                minTimeout: 100,
+                url: `https://${INTAKE_HOST}/${INTAKE_PATH}`,
+                method: 'POST',
+                type: 'json',
+                getData: async () => {
+                    const data = {
+                        ddsource: `@datadog/${bundler.name}-plugin`,
+                        env: context.env,
+                        message,
+                        service: 'build-plugins',
+                        bundler,
+                        plugins: context.pluginNames,
+                        version: context.version,
+                        team: 'yoann',
+                        ...ctx,
+                    };
+                    return {
+                        data: JSON.stringify(data),
+                        headers: {
+                            'Content-Type': 'application/json',
+                        },
+                    };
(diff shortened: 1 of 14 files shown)


packages/plugins/analytics/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                getData: async () => {
+                    const data = {
+                        ddsource: `@datadog/${bundler.name}-plugin`,
+                        env: context.env,
+                        message,
+                        service: 'build-plugins',
+                        bundler,
+                        plugins: context.pluginNames,
+                        version: context.version,
+                        team: 'yoann',
```
Ayc0, line 43: shouldn't this be `context.team`? x)

packages/plugins/analytics/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                        ...ctx,
```
Ayc0, line 44: FYI: if `ctx` contains `service`, or `bundler`, or `plugins`, it'll override the fields you setup here. Idk if it's expected or not, but if it isn't, you may want to put `...ctx` as the 1st element of this object

packages/plugins/analytics/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                        data: JSON.stringify(data),
+                        headers: {
+                            'Content-Type': 'application/json',
+                        },
+                    };
+                },
+            });
+        } catch (e: any) {
+            // We don't want to break anything in case of error.
+            log.debug(`Could not submit data to Datadog: ${e.message}`);
```
Ayc0, line 56: `e.message` may not exist (like if we threw a string)

You may want to use `e: unknown` instead, and do something like `` log.debug(`Could not submit data to Datadog:`, e) ``

Ayc0 (approved):

packages/plugins/analytics/src/index.ts
yoannmoinet: there is no team right now, I am my own team :)

packages/plugins/analytics/src/index.ts
yoannmoinet: I'm not sure what's preferable, being able to override the defaults, or losing what you may pass as the context.

I'll rename it as `overrides` so it's clearer that you can bypass the defaults with it.

packages/plugins/analytics/src/index.ts
yoannmoinet: I'll use `e.toString()`.
I don't think `log()` supports multi params (yet).

packages/plugins/analytics/src/index.ts
yoannmoinet: I could use `language-foundations` I suppose.

packages/plugins/analytics/src/index.ts
```diff
@@ -0,0 +1,69 @@
...
+            return;
+        }
+
+        try {
+            const bundler = {
+                name: context.bundler.name,
+                version: context.bundler.version,
+            };
+
+            await doRequest({
```
elbywan, line 27: Wondering, is this worth awaiting? I'm not sure whether we want to delay the build if the network is slow, are there any downsides when logging asynchronously?

yoannmoinet: I guess it can be up to the user of `context.sendLog()` to decide to await or not.

So the function does await, but you don't have to when you call it.

If the build exits prematurely it may cancel the call too, in case it's not awaited.

elbywan: > I guess it can be up to the user of context.sendLog() to decide to await or not.

ah true 🤦

PR #138, opened 2025-02-18
[ci] Release process update

yoannmoinet:
### What and why?

Having a way to manually bump the versions without checking out `master` would be great.

### How?

- Add a new `bump` action.
  1. run `yarn version:all <patch|minor|major>` based on the input of the workflow dispatch.
  2. commit the diff in `v<new_version>`.
  3. tag the commit with `v<new_version>`. 

[Successful run.](https://github.com/DataDog/build-plugins/actions/runs/13418889087/job/37486473979)

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -1,0 +1,43 @@
+name: Bump the versions
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'What to bump?'
+        required: true
+        default: 'patch'
+        type: choice
+        options:
+        - 'patch'
+        - 'minor'
+        - 'major'
+
+
+jobs:
+  bump:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # Needed for "yarn version" to work.
+        token: ${{ secrets.DD_INFRA_BOT_GH_ACTIONS_TOKEN }} # Needed for the push at the end of the action.
+    - uses: actions/setup-node@v4
+      with:
+        node-version-file: 'package.json'
+
+    - run: yarn install
+    - run: yarn version:all ${{ inputs.version }}
+    - run: git add .
+    - run: |
+        TAG_NAME="v$(yarn workspace @datadog/webpack-plugin info @datadog/webpack-plugin --json | jq -r '.children.Version')"
+        GIT_AUTHOR_NAME='dd-infra-bot',
+        GIT_COMMITTER_NAME='dd-infra-bot',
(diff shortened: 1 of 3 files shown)


.github/workflows/bump.yaml
```diff
@@ -0,0 +1,31 @@
...
+      with:
+        node-version-file: 'package.json'
+
+    - run: yarn version:all ${{ inputs.version }}
+    - run: git add .
+    - run: VERSION=`yarn workspace @datadog/webpack-plugin info @datadog/webpack-plugin --json | jq -r '.children.Version'`
+    - run: TAG_NAME="v$VERSION"
+    - run: git commit -m $TAG_NAME
+    - run: git tag -a $TAG_NAME -m $TAG_NAME
+    - run: git push --follow-tags
```
yoannmoinet, line 31: I think this might not work without a proper token.
But I need to merge this first in order to be able to actually debug it.

elbywan: > But I need to merge this first in order to be able to actually debug it.

FWIW it is possible to trigger a workflow dispatch from a branch using the gh tool (`gh workflow run workflow-name --ref branch-name`)

yoannmoinet: 😞 

```bash
╰─❯ gh workflow run bump.yaml --ref origin/yoann/release-process-update -f version=patch

HTTP 404: Not Found (https://api.github.com/repos/DataDog/build-plugins/actions/workflows/bump.yaml)
```

package.json
```diff
@@ -23,11 +23,12 @@
         "dev": "yarn cli prepare-link && yarn watch:all; yarn cli prepare-link --revert",
         "format": "yarn lint --fix",
         "lint": "eslint ./packages/**/*.{ts,js} --quiet",
-        "loop": "yarn workspaces foreach -Apti --include \"@datadog/*\" --exclude \"@datadog/build-plugins\"",
+        "loop-published": "yarn workspaces foreach -A --include \"@datadog/*\" --exclude \"@datadog/build-plugins\"",
```
elbywan, line 26: you might want to use the `-t` flag (topological) to publish dependencies first, in case something goes wrong

elbywan (approved):

package.json
yoannmoinet: I'm not sure it's really necessary.

It's not publishing anything, it's a simple loop helper over the workspaces that we publish (eg. `@datadog/*-plugin`).

I wanted (needed) a flat, sequential loop in order to (in this case) bump the versions in their `package.json`.

For the needs of topological, I'd use the `loop` helper, that adds the `-pti` flags to it (to also make it parallel).

PR #145, opened 2025-02-27
[web-ui] Quick fixes for web-ui

yoannmoinet:
### What and why?

Related to web-ui's integration.
Upgrading it to latest versions.
Some small fixes where necessary.

### How?

- esbuild: stub file was being removed too soon and sub builds would fail on resolution.
  - instead of creating the file in the outDir, we now use a tmpdir, and we don't delete it.
- rspack: removal of stub file wasn't happening in case of failed build (using `shutdown` instead of `done` now).
  - using `shutdown` async would randomly crash too, so made it sync.
- rspack: re-organised the code to be all at the same place.
- `yarn cli prepare-link` wasn't reverting to normal after a slight drift from `package.json` latest definition.
- small simplification of `runBundlers` by removing the `isolateModules` usage.
- fix the build-report to account for externals better and report them.
- move a bunch of helpers in the core package.
- add tests to some missing helpers.
- fix the `env` var to correctly fallback to `development`.

### Follow-ups

Write a test that would catch the esbuild issue better.
The current test might be running too fast and doesn't reproduce the edge case well enough.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,7 +90,7 @@ jobs:
         key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
 
     - uses: actions/upload-artifact@v4
-      if: ${{ !cancelled() && failure() }}
+      if: ${{ failure() }}
       with:
         name: playwright
         path: |
--- a/global.d.ts
+++ b/global.d.ts
@@ -2,13 +2,51 @@ import type { Env } from '@dd/core/types';
 
 declare global {
     namespace NodeJS {
-        interface ProcessEnv {
-            [key: string]: string | undefined;
+        interface ProcessEnv extends NodeJS.ProcessEnv {
+            /**
+             * To use when building the plugins with `yarn build:all`.
+             *
+             * If passed, the build will include our rollup's plugin to the build.
+             *
+             * Be sure to have build rollup's plugin with `yarn workspace @datadog/rollup-plugin build`.
+             */
+            ADD_BUILD_PLUGINS?: '1';
+            /**
+             * The environment in which the plugins will execute.
+             *
+             * For instance, we only submit logs to Datadog when the environment is `production`.
+             */
             BUILD_PLUGINS_ENV?: Env;
-            NO_CLEANUP?: '1';
-            NEED_BUILD?: '1';
-            REQUESTED_BUNDLERS?: string;
+            /**
+             * Defined in github actions when running in CI.
+             */
+            CI?: '1';
+            /**
+             * Defined in github actions when running in CI.
+             *
+             * The commit SHA that triggered the workflow.
+             */
+            GITHUB_SHA?: string;
+            /**
+             * Run jest in silent mode.
+             */
             JEST_SILENT?: '1';
+            /**
+             * To also build the plugins before running the tests when using `yarn test:unit`.
+             */
+            NEED_BUILD?: '1';
+            /**
+             * To skip the cleanup of the temporary working dirs where we build `runBundlers()`.
+             */
+            NO_CLEANUP?: '1';
+            /**
+             * The list of bundlers to use in our tests.
+             */
+            REQUESTED_BUNDLERS?: string;
+            /**
+             * Defined by yarn and targets the root of the project.
+             */
+            PROJECT_CWD?: string;
         }
     }
 }
--- a/global.d.ts
+++ b/global.d.ts
@@ -4,14 +4,6 @@ declare global {
     namespace NodeJS {
         interface ProcessEnv extends NodeJS.ProcessEnv {
             /**
-             * To use when building the plugins with `yarn build:all`.
-             *
-             * If passed, the build will include our rollup's plugin to the build.
-             *
-             * Be sure to have build rollup's plugin with `yarn workspace @datadog/rollup-plugin build`.
-             */
-            ADD_BUILD_PLUGINS?: '1';
-            /**
              * The environment in which the plugins will execute.
              *
              * For instance, we only submit logs to Datadog when the environment is `production`.
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
         "loop-published": "yarn workspaces foreach -A --include \"@datadog/*\" --exclude \"@datadog/build-plugins\"",
         "loop": "yarn loop-published -pti",
         "oss": "yarn cli oss -d packages -l mit",
+        "playwright": "yarn workspace @dd/tests playwright",
         "publish:all": "yarn loop --no-private npm publish",
         "typecheck:all": "yarn workspaces foreach -Apti run typecheck",
         "version:all": "yarn loop-published version ${0} --immediate",
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -217,6 +217,9 @@ export const isInjectionFile = (filename: string) => filename.includes(INJECTED_
 export const rm = async (dir: string) => {
     return fsp.rm(dir, { force: true, maxRetries: 3, recursive: true });
 };
+export const rmSync = async (dir: string) => {
+    return fs.rmSync(dir, { force: true, maxRetries: 3, recursive: true });
+};
 
 // Mkdir recursively.
 export const mkdir = async (dir: string) => {
--- a/packages/core/src/helpers.ts
+++ b/packages/core/src/helpers.ts
@@ -16,8 +16,10 @@ import type {
     BundlerFullName,
     Entry,
     File,
+    GetCustomPlugins,
     GlobalContext,
     Input,
+    IterableElement,
     Logger,
     Output,
     RequestOpts,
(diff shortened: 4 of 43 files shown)


packages/core/src/helpers.ts
```diff
@@ -212,11 +218,18 @@ export const truncateString = (
...
+// From a bundler's name, is it part of the "xpack" family?
+export const isXpack = (bundlerName: BundlerFullName) =>
+    ['rspack', 'webpack4', 'webpack5', 'webpack'].includes(bundlerName);
+
 // Replacing fs-extra with local helpers.
 // Delete folders recursively.
 export const rm = async (dir: string) => {
     return fsp.rm(dir, { force: true, maxRetries: 3, recursive: true });
 };
+export const rmSync = async (dir: string) => {
```
jakub-g, line 230:
```suggestion
export const rmSync = (dir: string) => {
```

jakub-g: (maybe some usages to update with `await` to be removed? though unlikely)

packages/factory/src/helpers.ts
```diff
@@ -118,7 +118,9 @@ export const getContext = ({
         },
     };
 
-    const passedEnv: Env = (process.env.BUILD_PLUGINS_ENV as Env) || 'development';
+    // Use "production" if there is no env passed.
+    const passedEnv: Env = (process.env.BUILD_PLUGINS_ENV as Env) || 'production';
+    // Fallback to "development" if the passed env is wrong.
```
jakub-g, line 123: Just curious, what's the difference between development and production here? Log verbosity? Something more?

yoannmoinet: Right now, the main diff is logging to Datadog in the analytics plugin, only done in `production`.

About this change specifically: 
Before, it was always falling back to `development` when not defined.
And I forgot to explicitly define it when bundling and publishing it.
Resulting with a `development` env in actual production, so got me thinking that maybe the better default is `production`.
Jury's still out though.

packages/core/src/helpers.ts
yoannmoinet: Oh, nice catch !

.github/workflows/ci.yaml
```diff
@@ -90,7 +90,7 @@ jobs:
         key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
 
     - uses: actions/upload-artifact@v4
-      if: ${{ !cancelled() && failure() }}
+      if: ${{ failure() }}
```
yoannmoinet, line 93: This wasn't triggering when needed.

global.d.ts
yoannmoinet, line 1: Mostly about documentation, a little bit about typing.

package.json
```diff
@@ -26,6 +26,7 @@
         "loop-published": "yarn workspaces foreach -A --include \"@datadog/*\" --exclude \"@datadog/build-plugins\"",
         "loop": "yarn loop-published -pti",
         "oss": "yarn cli oss -d packages -l mit",
+        "playwright": "yarn workspace @dd/tests playwright",
```
yoannmoinet, line 29: Just an alias.

packages/core/src/helpers.ts
```diff
@@ -36,9 +39,12 @@ export const formatDuration = (duration: number) => {
...
     const milliseconds = d.getUTCMilliseconds();
-    return `${days ? `${days}d ` : ''}${hours ? `${hours}h ` : ''}${minutes ? `${minutes}m ` : ''}${
-        seconds ? `${seconds}s ` : ''
-    }${milliseconds ? `${milliseconds}ms` : ''}`.trim();
+    const timeString =
+        `${days ? `${days}d ` : ''}${hours ? `${hours}h ` : ''}${minutes ? `${minutes}m ` : ''}${
+            seconds ? `${seconds}s` : ''
+        }`.trim();
+    // Split here so we can show 0ms in case we have a duration of 0.
+    return `${timeString}${!timeString || milliseconds ? ` ${milliseconds}ms` : ''}`.trim();
```
yoannmoinet, line 47: It was printing `0ms` all the time.
So I fixed it and added tests for it.

packages/core/src/helpers.ts
```diff
@@ -411,3 +424,145 @@ export const unserializeBuildReport = (report: SerializedBuildReport): BuildRepo
         outputs,
     };
 };
+
```
yoannmoinet, line 427: After this, it's just moving helpers to a more central place.

packages/plugins/build-report/src/esbuild.ts
yoannmoinet, line 1: There was an issue with external dependencies.
Either ignored, missed or broken.
These changes in the `plugins/build-report` are about supporting external dependencies in all the bundlers.

Also added this to the tests.

packages/plugins/build-report/src/helpers.ts
```diff
@@ -114,7 +101,7 @@ export const cleanName = (context: GlobalContext, filepath: string) => {
             .split('!')
             .pop()!
             // Remove outDir's path.
-            .replace(context.bundler.outDir, '')
+            .replace(getAbsolutePath(context.cwd, context.bundler.outDir), '')
```
yoannmoinet, line 104: This is coming from a pretty deep issue with rollup and vite.

Commenting about it here as it's the first place it's visible.

--- 

Rollup doesn't really have a notion of `cwd`, or, it's always using `process.cwd()`.

In our `bundler-plugin` we try to streamline the `cwd` used in the build, so for rollup (and vite) it's a bit more complicated to actually compute, especially impactful when the tests' builds are executed from a place (process.cwd()) and the files are living in a separate place (temporary `workingDir`).

It breaks this computation, which breaks some other things, mostly formatting of outputs.

So now, instead of relying on `process.cwd()`, which is still there as the initial default, and the nearest common directory between outDir, cwd and entry files. 
We first try to find the highest `package.json` file from the `outDir`, and only fallback to the previous process if we could not find it.

It results in a more reliable way of computing a somewhat "accurate" cwd.

packages/plugins/build-report/src/xpack.ts
yoannmoinet, line 1: Supporting externals.

packages/plugins/bundler-report/src/index.ts
```diff
@@ -2,39 +2,27 @@
...
-
-    for (let i = 0; i < minLength; i++) {
-        // We use the first path as our basis.
-        const component = splitPaths[0][i];
-        if (splitPaths.every((parts) => parts[i] === component)) {
-            commonParts.push(component);
-        } else {
-            break;
-        }
+// Compute the CWD based on a list of directories and the outDir.
```
yoannmoinet, line 15: cf: https://github.com/DataDog/build-plugins/pull/145/files#r1991128004

For a more detailed explanation.

packages/plugins/bundler-report/src/index.ts
```diff
@@ -96,12 +88,14 @@ export const getBundlerReportPlugins = (context: GlobalContext): PluginOptions[]
...
+                    const outputOptions = Array.isArray(options.output)
+                        ? options.output
+                        : [options.output];
+                    for (const output of outputOptions) {
+                        handleOutputOptions(output);
+                    }
                 }
             },
-            outputOptions(options) {
-                handleOutputOptions(options);
```
yoannmoinet, line 103: This (and above) resulted in a double execution.

It seems that `outputOptions` is just a hook called for EACH output option (it can be an array) passed to the configuration.

Where `options` is the "raw" full option object.

packages/plugins/injection/src/esbuild.ts
yoannmoinet, line 1: For esbuild we're now writing the stub injection file into a temporary dir.

So we don't have to delete it after the build, and sub-builds won't fail when trying to resolve it without our resolver.

It happens because we mutate the `build.initialOptions.inject` with a "virtual" file that needs to be physically present in the file system because esbuild seems to try to resolve it when validating its options/entries, and only then our loader/resolver can take over.

Some other plugins (out of our scope/ecosystem) would implement sub builds, like for workers, inline conversions, etc... re-using the `build.initialOptions` that we mutated with an `inject` entry that is "virtual", but they would not re-use the `plugins` were our loader/resolver are.

Resulting in a resolution error on the file.

Now, the file always exists, no more errors.

packages/plugins/injection/src/index.ts
yoannmoinet, line 1: Cleaning, factorising and "only do what's necessary" update.

Some hooks were not universally necessary, `resolveId`, `loadInclude`, `load`.
So I removed them from here, and implemented the necessary steps in each bundler (where necessary).

Also grouped particularities together (for `xpack` for instance).

packages/plugins/injection/src/xpack.ts
yoannmoinet, line 1: Same kind of issue than the `esbuild` counterpart.

We need a stub file in the file system for `xpack`.

But it can't be in a temporary directory like `esbuild` because `xpack` won't resolve it correctly out of its context.

Also, using the `done` hook to delete the file wasn't correct, as it was only executed on success.

BUT, using `shutdown` in async mode with `rspack` would randomly fail.
AND webpack4 doesn't offer this hook, so we fallback to both `done` and `failed`.

Cross bundlers implementations is so cool.

packages/tests/jest.config.js
yoannmoinet, line 1: This was an issue a long time ago.
It started to break some mocks (especially with `memfs` and wasn't actually necessary.

packages/tests/playwright.config.ts
```diff
@@ -28,7 +28,8 @@ export default defineConfig<TestOptions>({
         bundlers: getRequestedBundlers(),
         trace: 'retain-on-failure',
     },
-    timeout: 5_000,
+    globalTimeout: process.env.CI ? 20 * 60 * 1000 : undefined,
+    timeout: 60_000,
```
yoannmoinet, line 32: Giving more space to playwright.

packages/tests/src/_jest/fixtures/hard_project/package.json
```diff
@@ -6,10 +6,5 @@
     "packageManager": "yarn@4.2.1",
     "dependencies": {
         "chalk": "2.3.1"
-    },
-    "devDependencies": {
-        "react": "19.0.0",
-        "react-dom": "19.0.0",
-        "react-router-dom": "6.28.0"
```
yoannmoinet, line 13: Not sure why it was there, probably some legacy left overs from the rum plugin's implementation.

packages/tests/src/_jest/helpers/runBundlers.ts
```diff
@@ -223,28 +223,29 @@ export const runBundlers = async (
...
     const bundlersToRun = BUNDLERS.filter(
         (bundler) => !bundlers || bundlers.includes(bundler.name),
     );
 
     const workingDir = await prepareWorkingDir(seed);
 
+    if (NO_CLEANUP) {
+        console.log(`[NO_CLEANUP] Working directory: ${workingDir}`);
+    }
+
```
yoannmoinet, line 237: Pretty useful log to investigate the directory when you use `NO_CLEANUP=1`.

packages/tests/src/unit/factory/helpers.test.ts
yoannmoinet, line 1: Update to cover for the rollup/vite's `cwd` compute update.

packages/tests/src/unit/plugins/build-report/index.test.ts
yoannmoinet, line 1: Update to cover the `external` support.

packages/tests/src/unit/tools/src/commands/create-plugin/index.test.ts
yoannmoinet, line 1: Same as `resetModules`, doesn't seem necessary anymore.

packages/tools/src/commands/prepare-link/index.ts
```diff
@@ -43,7 +43,11 @@ class PrepareLink extends Command {
                 const pkgJsonPath = path.resolve(ROOT, pkg.location, 'package.json');
                 const pkgJson = require(pkgJsonPath);
                 if (this.revert) {
-                    pkgJson.exports = { '.': './src/index.ts' };
+                    pkgJson.exports = {
+                        './dist/src': './dist/src/index.js',
+                        './dist/src/*': './dist/src/*',
+                        '.': './src/index.ts',
+                    };
```
yoannmoinet, line 50: Coming from the streamlining update on `./published/**/package.json`.

packages/tools/src/helpers.ts
yoannmoinet, line 1: Just a move.

packages/tools/src/rollupConfig.mjs
yoannmoinet, line 1: Better typechecking.

It was actually useless without the `@ts-check` 😆

global.d.ts
yoannmoinet, line 1: Mostly about documentation, a little bit about typing.

packages/core/src/helpers.ts
yoannmoinet, line 47: It was printing 0ms all the time.
So I fixed it and added tests for it.

packages/plugins/build-report/src/esbuild.ts
yoannmoinet, line 1: There was an issue with external dependencies.
Either ignored, missed or broken.
These changes in the plugins/build-report/* are about supporting external dependencies in all the bundlers.

Also added this to the tests.

elbywan (approved):

jakub-g (approved):

PR #148, opened 2025-03-21
[build-report] Optimise build-report to prevent memory hog over long running builds

yoannmoinet:
### What and why?

Running a devserver with webpack/rspack would often (always) end up with a `RangerError: Maximum call stack size exceeded` error.

image

### How?

The `build-report` was filling arrays over and over, with each build increasing their size.

Now the `build-report` cleans between builds, and keep state in a more efficient way (using `Set` and `Map`).

The indexation of modules doesn't log warnings anymore, we mitigate the previous duplication issues by merging differences into the previously indexed module.

Testing covers these changes and remain ✅


packages/plugins/build-report/src/xpack.ts
```diff
@@ -436,11 +470,14 @@ export const getXpackPlugin =
...
                         continue;
                     }
-
-                    if (outputFound.type !== 'map' && !entryOutputs.includes(outputFound)) {
-                        entryOutputs.push(outputFound);
+                    if (outputFound.type !== 'map' && !entryOutputs.has(outputFound.name)) {
+                        entryOutputs.set(outputFound.name, outputFound);
                         // We know it's not a map, so we cast it.
-                        entryInputs.push(...(outputFound.inputs as Input[]));
+                        for (const input of outputFound.inputs as Input[]) {
```
Ayc0, line 476: why is the cast necessary here?

Ayc0 (approved):

packages/plugins/build-report/src/xpack.ts
yoannmoinet: In the case of sourcemaps, `file.inputs` can be `Outputs` as sourcemaps are built out of `Output` type of files.

So since I'm checking it's not a sourcemaps earlier, I'm casting it to make the rest of code type error free.

Very much looking for a better way.

elbywan (approved):

PR #165, opened 2025-05-16
[chore] Update rspack to 1.3.9

yoannmoinet:
### What and why?

Because we're far behind and `shutdown` hook random bug is making tests flaky.

### How?

Update rspack to latest.


Ayc0 (approved):

elbywan (approved):

packages/tools/package.json
```diff
@@ -49,7 +49,7 @@
         "@inquirer/select": "2.3.3",
         "@rollup/plugin-commonjs": "28.0.1",
         "@rollup/plugin-node-resolve": "15.3.0",
-        "@rspack/core": "1.1.2",
+        "@rspack/core": "1.3.10",
```
Ayc0, line 52: Shouldn't this be a peer dependency btw?

Same for esbuild & waebpack?

Ayc0: Related, if this is for internal, shouldn't we bump the requirement on rspack to at least the 1.3.10 in the peerDeps (as it'll fail for older versions)

yoannmoinet: It's internal only here.

I thought about bumping the requirements in the public ones, but I feel like it's a big restriction and higher the bar of entry for the plugin.
The consequences are not breaking anything, it's just when using a specific hook with rspack (`shutdown`) it "sometimes" fail (once every 20 or so builds).
It felt low risk to really justify such bump in requirement.
I'll revise if I get reports over this though.

Ayc0: Noted, thanks!

PR #166, opened 2025-05-16
[feat] Add metadata configuration

yoannmoinet:
### What and why?

We need to be able to identify builds in case we may have multiple ones in the same process.

### How?

Add `metadata.name` configuration, so the user can better identify their builds on Datadog's side.

We use this information in logs, metrics and spans.


Ayc0 (approved):
LGTM

PR #170, opened 2025-05-16
[chore] Batched updates

yoannmoinet:
Relies on:

- https://github.com/DataDog/build-plugins/pull/172

### What and why?

Some small updates that I batched together.

### How?

Check list of commits for more details.


packages/tools/src/helpers.ts
```diff
@@ -196,6 +196,7 @@ export const getSupportedBundlers = (getPlugins: GetPlugins) => {
         },
         context: getContext({
             // We don't care, this is a hack.
```
Ayc0, line 198: x)

Ayc0 (approved):

packages/tools/src/helpers.ts
yoannmoinet: 😄 brutally honest
(also, this is internal only)

PR #171, opened 2025-05-16
[feat] Add true-end plugin

yoannmoinet:
### What and why?

Not all bundlers report the same `buildEnd` or even `writeBundle` hooks.

- `rollup` will call `writeBundle` as many times as there are outputs.
- `esbuild` has a pretty inacurate `buildEnd`.
- ...

### How?

Implement a more accurate `trueEnd` hook.

Had to make both `syncTrueEnd` and `asyncTrueEnd` because if your hook handler is asynchronous it can't execute at the same time as if it was synchronous, depending on the bundler...

I added these details in the documentation:

```
It may execute sooner than `syncTrueEnd` in some contexts:

- `esbuild` will call `asyncTrueEnd` before `syncTrueEnd`.
    - We use `build.onDispose`, for the latest hook possible in the build. The issue is, it's synchronous only. So we have to use `build.onEnd` for the asynchronous `asyncTrueEnd`, but it's called well before `build.onDispose`.
- `webpack 4` will only call `syncTrueEnd` if the build has an error. All good otherwise.
```

In short, it's best to use `syncTrueEnd` if you can.


packages/factory/README.md
```diff
@@ -82,6 +83,13 @@ Most of the time they will interact via the global context.
 
 #### [📝 Full documentation ➡️](/packages/plugins/injection#readme)
 
+
+### True End
```
Ayc0, line 87: I find the name a bit weird (but I don't have a better suggestion :/)

what about `onFullBuild`?

Ayc0 (approved):

packages/factory/README.md
yoannmoinet: I find `onFullBuild` a bit misleading too, for instance in `rollup` the concept of `buildEnd` happens before the `writeBundle`, as-in, rollup (and vite) separate both the build from the write process.

Can you elaborate on what you find weird with "true end"?
Maybe we could go with just `end`?

Ayc0: "true" doesn't really have a definition. And it sounds a bit more like something you'd say when speaking to someone vs something defined in a spec / technical document

yoannmoinet: After some discussion offline, it's difficult to find a term that would really define this hook, so for now, until we get better consensus, we'll stick with `true-end`.

PR #174, opened 2025-05-20
[chore] Fix own monitoring

yoannmoinet:
### What and why?

This PR includes a collection of improvements to fix and enhance our own monitoring infrastructure. As we're using our plugins to monitor our own builds, we needed to make several improvements to ensure accurate reporting and better data collection.

The basic build was actually broken, so there were no reporting anywhere 🤦 

### How?

- **Fix Rspack Version Reporting**: Use `bundler.rspackVersion` to properly report Rspack version instead of the underlying `webpack` version it reports at `bundler.version`.
- **Added Metrics Sent Tracking**: Add a `metrics.count` metric to track how many metrics are being sent to Datadog in each build.
- **Improved Filename Cleaning**: Created a `removeCommonPrefix` utility function to better clean file names used in our reports, resulting in cleaner and shorter naming in reports
- **Enhanced Telemetry Configuration**: 
  - Updated tags for better service identification and filtering
  - Added sorting to metrics names for more consistent logging output
- **Fix Build Configuration**: Improved propagation of build options to helper functions to ensure consistent behavior throughout the build process


packages/factory/src/index.ts
```diff
@@ -76,7 +76,7 @@ export const buildPluginFactory = ({
         const context: GlobalContext = getContext({
             start,
             options,
-            bundlerVersion: bundler.version || bundler.VERSION,
+            bundlerVersion: bundler.rspackVersion || bundler.version || bundler.VERSION,
```
Ayc0, line 79: why are we using 3 tags here?

packages/plugins/build-report/src/helpers.ts
```diff
@@ -82,6 +82,15 @@ export const cleanPath = (filepath: string) => {
     );
 };
 
+// From two file paths, remove the common path prefix.
+export const removeCommonPrefix = (filepath1: string, filepath2: string) => {
+    const commonPath = filepath1
+        .split('/')
+        .filter((part, index) => part === filepath2.split('/')[index])
+        .join('/');
```
Ayc0, line 90: Shouldn't we use `path.sep` instead of `/`?

Ayc0 (approved):

packages/factory/src/index.ts
yoannmoinet: It's to cover:

- `(webpack|esbuild|vite).version`
- `rollup.VERSION`
- `rspack.rspackVersion`

The `bundler` variable is the actual instance of the bundler we're currently using.

Ayc0: I see, the variable `bundler` is the library itself

packages/plugins/build-report/src/helpers.ts
```diff
@@ -82,6 +82,15 @@ export const cleanPath = (filepath: string) => {
...
+    return filepath1.replace(commonPath, '').replace(/^\//, '');
```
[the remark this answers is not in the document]
yoannmoinet: Ahah, very true, I'll update, plus I should use `path.sep` here too.

packages/factory/src/index.ts
Ayc0: Can you add a comment here? 🙏

yoannmoinet: Will do, good idea.

packages/plugins/build-report/src/helpers.ts
yoannmoinet: I'll remove the regex, I believe copilot added it here on its own.
I don't think it's necessary for this function to trim leading slashes.

Ayc0 (approved):

Ayc0 (approved):

Issue #175, opened 2025-05-26
`vite-plugin` build error


tywayne:
### Describe what happened
Error during build, attempt to access index `[0]` of an empty array.

```
dashboard:build: error during build:
dashboard:build: TypeError: Cannot read properties of undefined (reading '0')
dashboard:build:     at file:///vercel/path0/node_modules/.pnpm/@datadog+vite-plugin@2.5.3_vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0_/node_modules/@datadog/vite-plugin/dist/src/index.mjs:12:35598
dashboard:build:     at et (file:///vercel/path0/node_modules/.pnpm/@datadog+vite-plugin@2.5.3_vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0_/node_modules/@datadog/vite-plugin/dist/src/index.mjs:12:35686)
dashboard:build:     at config (file:///vercel/path0/node_modules/.pnpm/@datadog+vite-plugin@2.5.3_vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0_/node_modules/@datadog/vite-plugin/dist/src/index.mjs:12:37162)
dashboard:build:     at runConfigHook (file:///vercel/path0/node_modules/.pnpm/vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0/node_modules/vite/dist/node/chunks/dep-CevzF2vT.js:66905:25)
dashboard:build:     at resolveConfig (file:///vercel/path0/node_modules/.pnpm/vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0/node_modules/vite/dist/node/chunks/dep-CevzF2vT.js:66346:18)
dashboard:build:     at async build (file:///vercel/path0/node_modules/.pnpm/vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0/node_modules/vite/dist/node/chunks/dep-CevzF2vT.js:65405:18)
dashboard:build:     at async CAC.<anonymous> (file:///vercel/path0/node_modules/.pnpm/vite@5.4.15_@types+node@18.19.69_lightningcss@1.29.2_terser@5.37.0/node_modules/vite/dist/node/cli.js:829:5)
dashboard:build:  ELIFECYCLE  Command failed with exit code 1.
```

### Steps to reproduce the issue:
[...]

PR #176, opened 2025-05-26
[fix] Vite: Ensure `directories` is set before calling `getCwd`

tywayne:
### What and why?
Potential resolution to #175 

Vite builds fail due to a TypeError when attempting to set the cwd. The code as is is never inserting paths to the `directories` Set, causing the error down the line.

### How?
This PR
- moves the outDir check / set earlier in the flow
- calls `directories.add` with the derived path to ensure it is set
- this allows the `getCwd` call to not fail with the TypeError

### DIFF
--- a/packages/plugins/bundler-report/src/index.ts
+++ b/packages/plugins/bundler-report/src/index.ts
@@ -144,14 +144,15 @@ export const getBundlerReportPlugins: GetInternalPlugins = (arg: GetPluginsArg) 
         vite: {
             ...(rollupPlugin() as PluginOptions['vite']),
             config(config) {
+                if (config.build?.outDir) {
+                    context.bundler.outDir = config.build.outDir;
+                }
+                directories.add(context.bundler.outDir);
+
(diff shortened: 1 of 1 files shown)


yoannmoinet:
Hey 👋 
Thanks a ton for the PR and the detailed issue attached to it.
really appreciated.

I've looked into it a bit, mostly, your gut feeling feels right.
The `context.bundler.outDir` should be set earlier 100%.

Though, I believe `getNearestCommonDirectory` should still handle an empty array without having to "dodge" its lacking by adding the `outDir` to it.

I pushed some commits to reflect my thoughts.
Let me know what you think @tywayne, if that works for you, I'll merge and publish a new version for it.

yoannmoinet (approved):

yoannmoinet:
It was just published in [`v2.6.1`](https://github.com/DataDog/build-plugins/releases/tag/v2.6.1).

Thanks a ton for your help @tywayne!

PR #180, opened 2025-06-16
[RUM-10137] Add privacy plugin 

cy-moi:
### What and why?
The privacy plugin parses string literals at build time to extract strings on demand.
The extracted strings will be used for information masking in browser sdk at client side.

### How?
1. Created `rum-privacy-plugin` that contains interfaces for `js-instrumentation-wasm` and helper functions for `rum-browser-sdk`
2. Created e2e test for checking the helpers injection
3. Build the sub-builds to both `cjs` and `esm` format to support `exports`


packages/plugins/rum-privacy/src/types.ts
```diff
@@ -0,0 +1,23 @@
...
+
+export interface RumPrivacyOptions {
+    exclude?: RegExp[] | string[];
+    include?: RegExp[] | string[];
+    module?: 'cjs' | 'esm';
+    jsx?: boolean;
+    transformStrategy?: 'ast';
+    typescript?: boolean;
+    disabled?: boolean | undefined;
+}
```
yoannmoinet, line 13: I'm not sure I follow, are these settings the customer can tweak?
Do they actually need this?

packages/tools/src/rollupConfig.mjs
```diff
@@ -218,6 +218,17 @@ export const getSubBuilds = async (ddPlugin, packageJson, options) => {
...
+                    getOutput(
+                        packageJson,
+                        {
+                            format,
+                            sourcemap: false,
+                            plugins: [terser({ mangle: true })],
+                        },
+                        options,
+                    ),
+                );
```
yoannmoinet, line 231: 😎

packages/tests/src/e2e/privacyPlugin/privacyPlugin.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+                });
+            } else {
+                // fail the test
+                expect(true).toBe(false);
+            }
+        });
+
+        const button = page.getByTestId('load-script');
+        await button.click();
+        await page.waitForTimeout(2000);
```
yoannmoinet, line 119: 😬 can we instead wait for the resource to be requested/loaded?

Something like this:

```suggestion
        const scriptRequest = page.waitForResponse(`${testBaseUrl}/dist/display-string.js`);
        await button.click();
        expect((await scriptRequest).ok()).toBe(true);
```

This way, it will only move forward until it actually got the resource loaded.

packages/plugins/rum-privacy/src/index.test.ts
```diff
@@ -0,0 +1,56 @@
...
+                        jsx: undefined,
+                        transformStrategy: 'ast',
+                        typescript: undefined,
+                    },
+                }),
+            );
+            expect(transform.buildTransformOptions).toHaveBeenCalled();
+        });
+    });
+});
```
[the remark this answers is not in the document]
cy-moi: We do have integration tests in [`js-instrumentation-wasm`](https://github.com/DataDog/js-instrumentation-wasm/tree/main/tests), I would prefer to limit this testing there but could always add more projects in e2e tests.

packages/published/rollup-plugin/package.json
```diff
@@ -24,8 +24,6 @@
     "exports": {
         "./dist/src": "./dist/src/index.js",
         "./dist/src/*": "./dist/src/*",
-        "./dist-basic/src": "./dist-basic/src/index.js",
-        "./dist-basic/src/*": "./dist-basic/src/*",
```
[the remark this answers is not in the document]
yoannmoinet: Oh I think it's maybe `yarn cli prepare-link`!

packages/tests/src/e2e/privacyPlugin/privacyPlugin.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+        page,
+        bundler,
+        browserName,
+        suiteName,
+        devServerUrl,
+    }) => {
+        // skip for webpack4 because of dynamic import
+        if (bundler === 'webpack4') {
+            test.skip();
+        }
```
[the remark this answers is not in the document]
cy-moi: Webpack 4 does not support dynamic import, so the dynamically loading another script to with a button would not run here, which also means we don't need to monitor the updates either.

packages/tests/src/e2e/privacyPlugin/privacyPlugin.spec.ts
cy-moi: Tested with this, yet we still need to wait for the callbacks getting triggered here.

packages/tests/src/e2e/privacyPlugin/privacyPlugin.spec.ts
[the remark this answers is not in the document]
cy-moi: This test case is testing dynamic loading part, which is not applicable to webpack4. So I don't think we need to test it.

packages/tests/src/e2e/privacyPlugin/privacyPlugin.spec.ts
```diff
@@ -0,0 +1,104 @@
...
+                console.log('DD_ALLOW observer triggered');
+            });
+        });
+
+        const button = page.getByTestId('load-script');
+        const waitForLog = page.waitForEvent('console', (msg) =>
+            msg.text().includes('DD_ALLOW observer triggered'),
+        );
+        await button.click();
+        expect(await waitForLog).toBeTruthy();
```
yoannmoinet, line 102: Since you're gathering errors, you can do an assertion on them:
```suggestion
        expect(await waitForLog).toBeTruthy();
        expect(errors).toEqual([]);
```

Otherwise, remove the gathering mechanism.

yoannmoinet (approved):
LGTM, well done 👍
Just one remaining nitpick.

PR #183, opened 2025-06-17
[chore] add Claude Code infos

yoannmoinet:
### What and why?

Instrument Claude Code.

### How?

Add a `CLAUDE.md` and some commands, including self reflecting ones.


Ayc0 (approved):
I haven't read any of those MD, but I 100% trust you on them 👌

yoannmoinet:
> I haven't read any of those MD, but I 100% trust you on them 👌

Yeah, it's not for us to read anyways.
And I added a self reflecting command that will help improve them with time and usage.
So they will change the more we use them.

Issue #179, opened 2025-06-14
[Vite plugin] Incorrect sourcemap paths


TmLev:
### Describe what happened

`datadogVitePlugin` produces incorrect paths for minified assets, e.g. `/assets/home/runner/<REDACTED>/assets/DataGrid-DrStp6xt.js` (taken from "Uploaded Debug Symbols" hereinafter).

### Steps to reproduce the issue:

I took the following Vite config:

```ts
export default defineConfig({
  plugins: [
    datadogVitePlugin({
      errorTracking: {
        sourcemaps: {
          minifiedPathPrefix: "/assets/",
        },
      },
    }),

    // Other plugins omitted for brevity.
  ],

  build: {
    sourcemap: true,
    outDir: "./dist",
  },
});
```

And run `vite build --mode production`.

### Expected behaviour:

Sourcemaps should be associated with paths starting with `minifiedPathPrefix` which is `/assets/` in my case.

It even works when I use the following CLI command:
```shell
        `DATADOG_SITE=${VITE_DATADOG_SITE} pnpm exec datadog-ci sourcemaps upload ./dist/assets ` +
          `--service ${VITE_DATADOG_SERVICE} ` +
          `--release-version ${currentVersion} ` +
          `--minified-path-prefix /assets`
```

Final path: `/assets/DataGrid-HKXszdMm.js`

### Actual behaviour:

Sourcemaps get weird filesystem paths instead, e.g. 
```
/assets/home/runner/<REDACTED>/assets/DataGrid-DrStp6xt.js
```

### Additional context

I'm using Github Actions to deploy my Vite+React project (so the `/home/runner` part of the path comes from that).

OS version: `ubuntu-latest`
Node version: 20
Vite version: 5.4.14
`@datadog/vite-plugin` version: 2.6.1

An explanation of what might cause the bug: incorrect handling of `ourDir` and filepaths.
An explanation of how it can be fixed: trace the way final minified path gets assembled.

#### Examples of other `minifiedPathPrefix` & `outDir` combinations

I've tried using FQDN for the `minifiedPathPrefix`:
```ts
export default defineConfig({
  plugins: [
    datadogVitePlugin({
      errorTracking: {
        sourcemaps: {
          minifiedPathPrefix: "https://full-domain.com/assets/",
        },
      },
    }),
  ],

  build: {
    sourcemap: true,
    outDir: "./dist",
  },
});


[...]

PR #186, opened 2025-06-30
[fix] Vite outDir for sourcemaps upload (take 2)

yoannmoinet:
### What and why?

Fixing https://github.com/DataDog/build-plugins/issues/179 again, apparently #182 didn't fully fix it.

### How?

I think the uncertainty and flimsiness is coming from the internal `bundler-report` where `bundler.outDir` wouldn't be totally normalised across bundlers (sometimes relative, absolute, leading slashes, ...).

So I totally reworked this section to be normalised as an absolute path.
Added a ton of tests in both this section and the sourcemaps uploading.

### DIFF
--- a/packages/plugins/bundler-report/src/helpers/rollup.test.ts
+++ b/packages/plugins/bundler-report/src/helpers/rollup.test.ts
@@ -60,7 +60,7 @@ describe('Rollup Helpers', () => {
             },
             {
                 description: 'return undefined when no dir or file specified',
-                outputOptions: [{ format: 'esm' }, { format: 'cjs' }] as any,
+                outputOptions: [{ format: 'esm' }, { format: 'cjs' }],
                 expected: undefined,
             },
         ];
@@ -154,6 +154,7 @@ describe('Rollup Helpers', () => {
             const expectedErrors = shouldThrow ? [shouldThrow] : [];
 
             try {
+                // Rollup doesn't type InputOptions['output'], yet it exists.
                 const result = computeCwd(options as InputOptions);
                 results.push(result);
             } catch (error: any) {
--- a/packages/plugins/bundler-report/src/helpers/rollup.ts
+++ b/packages/plugins/bundler-report/src/helpers/rollup.ts
@@ -8,9 +8,8 @@ import type { InputOptions } from 'rollup';
 
 // Compute the CWD based on a list of directories.
 const getCwd = (dirs: Set<string>) => {
-    const dirsToUse: Set<string> = new Set();
+    const dirsToUse: Set<string> = new Set(dirs);
     for (const dir of dirs) {
-        dirsToUse.add(dir);
         const highestPackage = getHighestPackageJsonDir(dir);
         if (highestPackage && !dirs.has(highestPackage)) {
             dirsToUse.add(highestPackage);
@@ -19,11 +18,10 @@ const getCwd = (dirs: Set<string>) => {
 
     // Fall back to the nearest common directory.
     const nearestDir = getNearestCommonDirectory(Array.from(dirsToUse));
-    if (nearestDir !== path.sep) {
-        return nearestDir;
-    } else {
+    if (nearestDir === path.sep) {
         return undefined;
     }
+    return nearestDir;
 };
 
 export const getOutDirFromOutputs = (options: InputOptions) => {
@@ -59,11 +57,10 @@ export const computeCwd = (options: InputOptions) => {
               : [options.input];
 
         for (const input of normalizedInput) {
-            if (typeof input === 'string') {
-                directoriesForCwd.add(path.dirname(input));
-            } else {
+            if (typeof input !== 'string') {
                 throw new Error('Invalid input type');
             }
+            directoriesForCwd.add(path.dirname(input));
         }
     }
 


packages/plugins/bundler-report/src/index.ts
```diff
@@ -1,42 +1,32 @@
...
     (context) => (compiler) => {
         context.bundler.rawConfig = compiler.options;
 
         if (compiler.options.output?.path) {
-            context.bundler.outDir = compiler.options.output.path;
+            // While webpack doesn't allow for non-absolute paths,
+            // rspack does, and will fallback to use process.cwd().
+            context.bundler.outDir = getAbsoluteOutDir(process.cwd(), compiler.options.output.path);
         }
         context.hook('bundlerReport', context.bundler);
```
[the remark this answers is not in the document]
yoannmoinet: It is 5 lines below.

packages/plugins/bundler-report/src/index.test.ts
```diff
@@ -2,66 +2,386 @@
...
+                // Remove warning about unset mode.
+                mode: 'none',
+                entry: {
+                    main: path.resolve(cwd, defaultEntry),
+                },
+                output: {
+                    // Webpack won't allow relative paths.
+                    path: path.resolve(cwd, 'dist-webpack4'),
+                },
+                // Need to use an helper to differentiat webpack5 from webpack4.
```
sethfowler-datadog, line 219: Nit:
```suggestion
                // Need to use an helper to differentiate webpack5 from webpack4.
```

packages/plugins/bundler-report/src/index.test.ts
```diff
@@ -2,66 +2,386 @@
...
+                // Remove warning about unset mode.
+                mode: 'none',
+                entry: {
+                    main: path.resolve(cwd, defaultEntry),
+                },
+                output: {
+                    // Webpack won't allow relative paths.
+                    path: path.resolve(cwd, 'dist-webpack5'),
+                },
+                // Need to use an helper to differentiat webpack5 from webpack4.
```
sethfowler-datadog, line 239: Nit:
```suggestion
                // Need to use an helper to differentiate webpack5 from webpack4.
```

packages/plugins/bundler-report/src/helpers/rollup.ts
```diff
@@ -0,0 +1,103 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import { getHighestPackageJsonDir, getNearestCommonDirectory } from '@dd/core/helpers/paths';
+import path from 'path';
+import type { InputOptions, OutputOptions, RollupOptions } from 'rollup';
+
+// Compute the CWD based on a list of directories.
+const getCwd = (dirs: Set<string>) => {
```
sethfowler-datadog, line 10: While the mechanical changes in this PR generally make sense, I wonder if there's an opportunity to make things clearer by adjusting the terminology a bit.

The concept of the "current working directory" of a process is well defined, and so it seems to me that the name of this function, and the other functions related to this task, are probably not quite right. We are actually computing something more like a "build root", right? It's just that this happens to correspond to the CWD for some bundlers, if I understand what's happening here correctly.

Perhaps the problem is broader than just these functions, in fact; should `GlobalContext` be exposing something like `buildRoot`, instead of `cwd`?

I'm not sure how much code this change would affect outside of the files that are already being changed in this PR, and I don't want to block your ability to merge. Happy to drop this suggestion if it doesn't make sense, or alternatively, happy to review a separate PR to tweak the terminology so you can get this one merged.

packages/plugins/bundler-report/src/helpers/rollup.ts
```diff
@@ -0,0 +1,103 @@
...
+              : [options.input];
+
+        for (const input of normalizedInput) {
+            if (typeof input === 'string') {
+                directoriesForCwd.add(path.dirname(input));
+            } else {
+                throw new Error('Invalid input type');
+            }
+        }
+    }
```
Ayc0, line 84: if there is no `input`, what do we do?

Ayc0: Shouldn't we do something like:

```js
if (!option.input) {
  throw new Error('input is required');
}
```

and then remove this if so that we'll always be in the happy path?

packages/plugins/bundler-report/src/helpers/rollup.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    const cwd = getCwd(directoriesForCwd);
+
+    if (cwd) {
+        return cwd;
+    }
+
+    // Fallbacks
+    return process.cwd();
```
Ayc0, line 102: I find it a bit weird that `getCwd()` doesn't auto does the `process.cwd()` fallback. Like you get it, but not fully

packages/plugins/bundler-report/src/helpers/rollup.ts
yoannmoinet: The `input` configuration isn't required by the bundler.
So we only use it to compute the build root if we have it.

Otherwise, we just continue our path and don't use it for our computation as the bundler will fall back to `process.cwd()` just like our script.

packages/plugins/bundler-report/src/helpers/rollup.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (!outDir) {
+        return '';
+    }
+
+    return path.isAbsolute(outDir) ? outDir : path.resolve(cwd, outDir);
+};
+
+export const getOutDirFromOutputs = (outputOptions: RollupOptions['output']) => {
+    if (!outputOptions) {
+        return '';
```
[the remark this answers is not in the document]
yoannmoinet: I removed the condition entirely 😄

packages/plugins/bundler-report/src/helpers/rollup.ts
yoannmoinet: 100% agreed, I'll change `cwd` to `buildRoot` in a different PR.
In that other PR, I'll also add support for the "multi outputs" of rollup and vite which kind of messes with everything else here...
This will probably be breaking enough to be included in the `v3` release.

packages/plugins/bundler-report/src/helpers/rollup.test.ts
```diff
@@ -0,0 +1,167 @@
...
+                expected: undefined,
+            },
+            {
+                description: 'return undefined for empty array',
+                outputOptions: [],
+                expected: undefined,
+            },
+            {
+                description: 'return undefined when no dir or file specified',
+                outputOptions: [{ format: 'esm' }, { format: 'cjs' }] as any,
```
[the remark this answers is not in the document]
yoannmoinet: I think it's because it's supposed to be `OutputOptions` but this one isn't type compliant without actual `dir` or `outfile`.
I'll double-check.

Ayc0 (approved):
My remaining comments are just suggestions for improvements, not required changes, so lgtm

Feel free to apply them if you want

PR #187, opened 2025-07-04
[rum-privacy] add sourcemaps and telemetry

cy-moi:
### What and why?
Improve the privacy plugin and update to the latest `js-instrument-wasm` library.
We want to add sourcemaps and telemetry to our privacy plugin.

### How?
Change `enforce: pre` to `post` for privacy plugin
Make sdk injection plugin optional when opt-in for rum plugin
Add `Typescript` end to end testing and support
Update transform options to make sourcemaps working
Add logs for observability


packages/plugins/rum/src/privacy/index.ts
```diff
@@ -12,38 +12,54 @@ import { PRIVACY_HELPERS_MODULE_ID, PLUGIN_NAME } from './constants';
...
+                const code = fs.readFileSync(privacyHelpersPath, 'utf8');
+                if (context.bundler.name === 'rollup') {
+                    // prepend AAAA to sourcemap
+                    const sourcemap = {
+                        version: 3,
+                        sources: [privacyHelpersPath],
+                        sourcesContent: [code],
+                        names: [],
+                        mappings: Array(code.split('\n').length).fill('AAAA').join(';'),
+                    };
```
sethfowler-datadog, line 59: What you are doing here is generating an identity source map, right? (i.e., a source map that maps every line to itself.) It's a bit weird that rollup needs this; usually rollup is the most well-behaved bundler! But if we must, we must.

Instead of the `// prepend AAAA to sourcemap` comment, which is true but doesn't clearly explain what's really happening, it'd be better to pull this code into a separate function called `generateIdentitySourceMap()` or something (to clarify _what_ it does) and then call it here with a comment explaining _why_ we need to do it. (Like "rollup doesn't generate any mappings for the privacy helpers in the output if we don't provide a source map", if that's the reason.) Otherwise, it'll be really hard to understand what's happening here later, especially for folks who haven't already spent time in the source map trenches.

packages/plugins/rum/src/privacy/transform.ts
```diff
@@ -22,8 +22,9 @@ export function buildTransformOptions(pluginOptions: PrivacyOptions): Instrument
         privacy: {
             addToDictionaryHelper: {
                 import: {
-                    module: PRIVACY_HELPERS_MODULE_ID,
-                    func: '$',
+                    cjsModule: pluginOptions.helpersModule ?? `${PRIVACY_HELPERS_MODULE_ID}.cjs`,
+                    esmModule: pluginOptions.helpersModule ?? `${PRIVACY_HELPERS_MODULE_ID}.mjs`,
```
sethfowler-datadog, line 26: Does `pluginOptions` need a `cjsHelpersModule` and an `esmHelpersModule`, since we needed that ourselves?

packages/plugins/rum/src/privacy/types.ts
```diff
@@ -7,6 +7,8 @@ import type { Assign } from '@dd/core/types';
 export interface PrivacyOptions {
     exclude?: RegExp[] | string[];
     include?: RegExp[] | string[];
+    helpersModule?: string;
+    globalFunc?: string;
```
sethfowler-datadog, line 11: The name `globalFunc` is pretty vague; it might be good to use a more specific name.

packages/plugins/rum/src/privacy/types.ts
```diff
@@ -16,5 +18,5 @@ export interface PrivacyOptions {
 
 export type PrivacyOptionsWithDefaults = Assign<
     PrivacyOptions,
-    Pick<Required<PrivacyOptions>, 'exclude' | 'include' | 'module' | 'transformStrategy'>
+    Pick<Required<PrivacyOptions>, 'exclude' | 'include' | 'transformStrategy'>
```
sethfowler-datadog, line 21: It doesn't need to block this PR, but before we ship this, we should clean up these options a bit. It looks like there's some obsolete stuff here; `transformStrategy` isn't accepted anymore by `@js-instrumentation-wasm`, for example, but the plugin is still accepting it.

packages/tests/src/e2e/privacyPluginTypescript/privacyPluginTypescript.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+            }
+        });
+
+        await userFlow(testBaseUrl, page, bundler);
+
+        const ddAllow = await page.evaluate(() => {
+            return (globalThis as any).$DD_ALLOW;
+        });
+
+        expect(ddAllow).toBeDefined();
```
sethfowler-datadog, line 82: It'd be great if we could assert something about the actual contents of `$DD_ALLOW`. I know the results may vary slightly depending on which bundler you use, so it might be hard to use `toEqual()`, but maybe we could check for some strings that we expect to be present.

sethfowler-datadog (approved):
Looks great! I posted a few nits below:

packages/plugins/rum/src/sdk.ts
```diff
@@ -25,6 +25,9 @@ export const getInjectionValue = (
     context: GlobalContext,
 ): InjectedValue => {
     const sdkOpts = options.sdk;
+    if (sdkOpts.disabled) {
+        return '';
+    }
```
yoannmoinet, line 30: Is this really necessary?

packages/plugins/rum/src/validate.test.ts
```diff
@@ -0,0 +1,77 @@
...
+            mockLogger,
+        );
+        options.privacy?.exclude.forEach((regex) => {
+            if (typeof regex === 'string') {
+                return;
+            }
+            expect(regex.test('.preval.js')).toBe(true);
+        });
+    });
+});
```
yoannmoinet, line 77: Could this be done simply using `createFilter` and a list of tests cases:

```ts
import { createFilter } from '@rollup/pluginutils';

describe('Privacy options', () => {
    let filter;
    const testCases = [
        { description: 'exclude .preval files', path: 'some-path', expected: false },
        // [...]
    ];

    beforeAll(() => {
        const pluginOptions = { ...defaultPluginOptions, rum: { privacy: {} } };
        const { include, exclude } = validateOptions(pluginOptions, mockLogger);
        filter = createFilter(include, exclude);
    });
    test.each(testCases)('Should $description', ({ path, expected }) => {
        expect(filter(path)).toBe(expected);
    });
});
```

packages/plugins/rum/src/validate.ts
```diff
@@ -56,6 +59,9 @@ export const validateSDKOptions = (options: Options): ToReturn<SDKOptionsWithDef
     const toReturn: ToReturn<SDKOptionsWithDefaults> = {
         errors: [],
     };
+    if (!validatedOptions.sdk || validatedOptions.sdk.disabled) {
+        return toReturn;
+    }
```
yoannmoinet, line 64: With this `if` as an early out, I think you can remove the one right below, `if (validatedOptions.sdk)`.

packages/plugins/rum/src/types.ts
```diff
@@ -19,39 +19,48 @@ export type RumOptions = {
...
+              | 'storeContextsAcrossPages'
+              | 'telemetrySampleRate'
+              | 'traceSampleRate'
+              | 'trackingConsent'
+              | 'trackLongTasks'
+              | 'trackResources'
+              | 'trackUserInteractions'
+              | 'trackViewsManually'
+          >
+      >;
```
yoannmoinet, line 63: Are these type changes still necessary?

packages/plugins/rum/src/validate.ts
```diff
@@ -122,5 +131,7 @@ export const validatePrivacyOptions = (options: Options): ToReturn<PrivacyOption
         };
     }
 
+    log.debug(`datadog-rum-privacy plugin options: ${JSON.stringify(toReturn.config)}`);
+
```
yoannmoinet, line 135: Remember to remove this one before merge 😄

packages/plugins/rum/src/validate.ts
```diff
@@ -101,18 +107,21 @@ export const validateSDKOptions = (options: Options): ToReturn<SDKOptionsWithDef
...
+): ToReturn<PrivacyOptionsWithDefaults> => {
     const validatedOptions: RumOptions = options[CONFIG_KEY] || {};
     const toReturn: ToReturn<PrivacyOptionsWithDefaults> = {
         errors: [],
     };
 
     if (validatedOptions.privacy) {
         const privacyWithDefault: PrivacyOptionsWithDefaults = {
-            exclude: [/\/node_modules\//, /\.preval\./],
+            exclude: DEFAULT_EXCLUDE,
```
yoannmoinet, line 121: Why the variable?
Also, can you add comments on the regexes you use, just being explicit about what's excluded and what's included.
This way, when someone reads your code, they can confirm the regex does what it's supposed to.

packages/tests/src/_playwright/helpers/buildProject.ts
```diff
@@ -3,31 +3,66 @@
...
 // Build a given project with a given bundler.
-const buildProject = async (bundler: BundlerFullName, cwd: string) => {
-    const plugin = allPlugins[bundler](fullConfig);
+const buildProject = async (
+    bundler: BundlerFullName,
+    cwd: string,
+    pluginConfigOverride?: Options,
+    buildConfigOverride?: BundlerConfig,
+) => {
+    const plugin = allPlugins[bundler](pluginConfigOverride || fullConfig);
```
yoannmoinet, line 22:
```suggestion
const buildProject = async (
    bundler: BundlerFullName,
    cwd: string,
    pluginConfig?: Options = fullConfig,
    buildConfigOverride?: BundlerConfig,
) => {
    const plugin = allPlugins[bundler](pluginConfig);
```

packages/tests/src/_playwright/helpers/buildProject.ts
```diff
@@ -3,31 +3,66 @@
...
+            // Use @rollup/plugin-typescript for Rollup and Vite
+            additionalPlugins.push(
+                typescript({
+                    tsconfig: path.resolve(cwd, 'tsconfig.json'),
+                }),
+            );
+        }
+        // ESBuild has built-in TypeScript support, no additional plugins needed
+    }
+
```
yoannmoinet, line 47: Can't this logic live at the test level?

As-in, if your test requires to override the build's configuration, it should also handle this part as well (adding the ts plugin to the build).

Can you also give me more details on why you need this `@rollup/plugin-typescript` plugin exactly and not re-use the `esbuild` plugin?

packages/tools/src/bundlers.ts
```diff
@@ -211,6 +214,18 @@ export const configXpack = (
...
+
+    // Add TypeScript support for webpack/rspack
+    if (hasTypeScriptEntries) {
+        baseConfig.resolve = {
+            extensions: ['.tsx', '.ts', '.js'],
+        };
+        baseConfig.module = {
+            rules: [{ test: /\.([cm]?ts|tsx)$/, loader: 'ts-loader' }],
+        };
+    }
```
yoannmoinet, line 226: Does this even need to be conditional?
Don't we always want to use `ts-loader` for any `ts` file we encounter? I don't thing we need `hasTypeScriptEntries`.

packages/tools/src/plugins.ts
```diff
@@ -32,11 +32,12 @@ export const fullConfig: Options = {
     },
     [RUM]: {
         sdk: {
+            disabled: false,
             applicationId: '123',
             clientToken: '123',
         },
         privacy: {
-            disabled: false,
+            disabled: true,
```
yoannmoinet, line 40: Why having this disabled in the `fullConfig`?

yoannmoinet (commented):
I'm not a fan of the addition to the `verifyProjectBuild`. 
It's very complex, I'll have a look, but I think this can be simplified.

packages/plugins/rum/src/validate.ts
cy-moi: This should include only `privacy` related configurations now. I'm trying to collect user configuration telemetries.

packages/tests/src/_playwright/helpers/buildProject.ts
cy-moi: I did try this, but this require changing the params for `verifyProjectBuild` even more, as we need to be able to accept different plugins for different bundlers. We are using `@rollup/plugin-typescript` to isolate issues of generated sourcemaps and avoid having two bundlers for the same build, per discussions in slack previously.

packages/plugins/rum/src/built/privacy-helpers.ts
```diff
@@ -9,7 +9,15 @@ globalAny.$DD_ALLOW = new Set();
...
-    newValues.forEach((value) => globalAny.$DD_ALLOW.add(value));
+    newValues.forEach((value) => {
+        if ((value as unknown as TemplateStringsArray).raw) {
+            (value as unknown as TemplateStringsArray).raw.forEach((raw) => {
+                globalAny.$DD_ALLOW.add(raw);
+            });
+        } else {
+            globalAny.$DD_ALLOW.add(value);
+        }
+    });
```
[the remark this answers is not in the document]
sethfowler-datadog: The code generated by `@datadog/js-instrumentation-wasm` was designed with this implementation in mind:
```suggestion
  if ((items as TemplateStringsArray).raw) {
    // We're being used as a template tag function. The invocation will look like this:
    //   const D = $('foo', $`bar${0}`, 'baz');
    // In this context, our only role is to extract the TemplateStringsArray array so that
    // the top-level call to $ can make use of it. So, we just need to return our first
    // argument.
    return items;
  }
```
In other words, if the property `raw` is present on `newValues`, this indicates that `$` is being used as a template tag function, and in this case we should just return the first argument and let the top-level invocation of `$` handle things. This feels conceptually simpler to me, since the top-level invocation of `$` in each module is the only place that strings are added to the dictionary; everything happens in a single loop. That said, you certainly don't have to do it that way; you could process the template quasis eagerly instead, and then skip over them in the loop for the top-level invocation of `$`. This requires two loops and slightly more complex logic, so I prefer the other approach, but it's up to you.

A less subjective matter: we should not be adding the values in `raw` to the dictionary. Those are the unescaped versions of the template quasis; we want the escaped versions, which are the versions that you get if you just iterate over the `TemplateStringsArray` like an array.

@cy-moi , you may find [the original implementation](https://github.com/DataDog/js-instrumentation-wasm/blob/main/tests/instrumentation-test-plugin/privacy-helpers/index.ts) helpful as a reference. Here's a summary of the logic in that version:

1. If `newValues` has a `raw` property, we're being used as a template tag function. Don't do any further processing; just return `newValues` as-is.
2. Otherwise, iterate over `newValues`. For each `value` in `newValues`:
  1. If `value` is a `string`, add it to the dictionary.
  2. Otherwise, treat `value` as an array of strings and add each string it contains to the dictionary.
3. Return `newValues`.

packages/plugins/rum/src/validate.ts
yoannmoinet: > I'm trying to collect user configuration telemetries.

🤔 this won't work the way you expect it to work then.

Here it will log for everyone (including the ones not enabling the privacy plugin) and it will only log locally to them, as in, we won't see it on our side.

cy-moi: Yes, I will add the `forward` key later (hence the merge :D). But I'm counting on it to be safe to log. Otherwise, I can remove this.

packages/plugins/rum/src/built/privacy-helpers.ts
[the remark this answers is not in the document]
cy-moi: Thanks for the explanation! This has been fixed.

packages/plugins/rum/src/validate.ts
```diff
@@ -56,71 +56,84 @@ export const validateSDKOptions = (options: Options): ToReturn<SDKOptionsWithDef
     const toReturn: ToReturn<SDKOptionsWithDefaults> = {
         errors: [],
     };
+    if (!validatedOptions.sdk) {
```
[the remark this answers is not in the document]
yoannmoinet: Oh no need to add it back.
I just missed that part :man_facepalming: 
All good then!

packages/plugins/injection/src/esbuild.ts
```diff
@@ -118,15 +118,19 @@ export const getEsbuildPlugin = (
...
+                    const data = await esbuild.transform(source, {
+                        loader: 'default',
+                        banner,
+                        footer,
+                    });
+
+                    // FIXME: Handle sourcemaps.
+                    await fsp.writeFile(output, data.code);
+                } catch (e) {
+                    // skip if the file doesn't exist
```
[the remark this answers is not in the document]
cy-moi: For example, when we have sub-builds in web-ui, the build entries are not actual entries and it may not exist when injecting. I'm not sure we should throw the error when the file does not exist, as it is not the entry file we want to inject anyway. I added comments and a log here instead.

packages/plugins/rum/src/index.test.ts
```diff
@@ -35,7 +35,7 @@ describe('RUM Plugin', () => {
         },
         {
             type: 'sdk',
-            config: { sdk: { applicationId: 'app-id' } },
+            config: { sdk: { applicationId: 'app-id', clientToken: '123' } },
```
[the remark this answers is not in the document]
cy-moi: There is no longer an issue after restoring to previous types. `clientToken` is not required. I restored this as well.

packages/plugins/rum/src/index.ts
```diff
@@ -55,12 +55,15 @@ export const getPlugins: GetPlugins = ({ options, context }) => {
         });
     }
 
-    if (validatedOptions.privacy) {
+    if (validatedOptions.privacy && !validatedOptions.privacy.disabled) {
```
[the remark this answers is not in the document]
cy-moi: I removed the `privacy.disabled` option as well, per bug bash feedback.

yoannmoinet (commented):
Small nits.
Waiting for the update on the esbuild skipping errors then I'll ✅

packages/plugins/injection/src/esbuild.ts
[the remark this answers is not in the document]
cy-moi: This should be fixed in the latest commit.

packages/plugins/rum/src/built/privacy-helpers.ts
```diff
@@ -7,14 +7,46 @@
...
+        // We're being used as a template tag function. The invocation will look like this:
+        //   const D = $('foo', $`bar${0}`, 'baz');
+        // In this context, our only role is to extract the TemplateStringsArray array so that
+        // the top-level call to $ can make use of it. So, we just need to return our first
+        // argument.
+        return newValues;
+    }
+
+    newValues
+        .flatMap((value) => value)
```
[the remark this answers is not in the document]
cy-moi: We do need the flat() here as we use `$` function also for string templates recursively.

PR #188, opened 2025-07-08
[chore] Claude configuration update

yoannmoinet:
### What and why?

I went too global with commands, let us only have things actually related to the repo.

### How?

Change commands to be more local-related.
Add new ones that are more useful.
Change configurations to also accept some safe bash commands.
Fix docs using new commands.


Ayc0 (approved):
AI? I'll trust you on the content

PR #191, opened 2025-07-22
[chore] Update bundlers

yoannmoinet:
## Update bundlers

Updates all supported bundlers to their latest versions and adjusts configurations.

### Bundler Updates

- **Webpack**: 5.92.1 → 5.100.2
- **ESBuild**: 0.24.0 → 0.25.8
- **Rollup**: 4.24.2 → 4.45.1
- **Rspack**: 1.3.10 → 1.4.9
- **Vite**: 6.3.4 → 6.3.5 (v7 not compatible due to ESM-only support)

### Other Changes

- Updated Jest configuration to follow `ts-node` guidance
- Updated license script to handle new packages


packages/tests/jest.config.ts
```diff
@@ -2,14 +2,22 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2019-Present Datadog, Inc.
 
-module.exports = {
+import { createJsWithTsPreset, type JestConfigWithTsJest } from 'ts-jest';
+
+const presetConfig = createJsWithTsPreset({
+    // ...options
```
jckr, line 8: is that something we want to change later ?

jckr (approved):

packages/tests/jest.config.ts
yoannmoinet: Good catch, I forgot to remove this after my copy pasting from their website 🤦
I'll remove it.

PR #202, opened 2025-08-05
[feat] Output plugin

yoannmoinet:
### What and why?

This PR adds the Output Plugin, providing the ability to export build reports, metrics, and bundler data to JSON files for analysis and/or monitoring.

### How?

- **New `output` plugin** (`packages/plugins/output/`) that exports various build artifacts to JSON files including build reports, metrics, timings, logs, errors, warnings, dependencies, and bundler data
- **Migrated debug functionality** from the separate debugFilesPlugin to be incorporated directly into the Output Plugin for better consolidation
- **Smaller fixes**
  - Fix `create-plugin` with updated templates.
  - Fixed some documentations.

### DIFF
--- a/packages/plugins/output/src/index.ts
+++ b/packages/plugins/output/src/index.ts
@@ -55,13 +55,17 @@ const getXpackPlugin =
 const getRollupPlugin = (
     write: (outputs: OutputBundle[]) => void,
 ): PluginOptions['rollup'] & PluginOptions['vite'] => {
-    const outputs: OutputBundle[] = [];
+    const outputs: Set<OutputBundle> = new Set();
     return {
+        buildStart() {
+            // Clear set on build start.
+            outputs.clear();
+        },
         writeBundle(opts, bundle) {
-            outputs.push(bundle);
+            outputs.add(bundle);
         },
         closeBundle() {
-            write(outputs);
+            write(Array.from(outputs));
         },
     };
 };
--- a/packages/plugins/output/src/index.ts
+++ b/packages/plugins/output/src/index.ts
@@ -94,7 +94,6 @@ export const getPlugins: GetPlugins = ({ options, context }) => {
         }
     };
 
-    // TODO: Add metrics report.
     return [
         {
             name: PLUGIN_NAME,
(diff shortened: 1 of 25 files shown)


yoannmoinet:
bugbot run

packages/plugins/output/src/index.ts
```diff
@@ -0,0 +1,146 @@
...
+                    start: serializedReport.start,
+                    end: serializedReport.end,
+                    duration: serializedReport.duration,
+                    writeDuration: serializedReport.writeDuration,
+                    entries: serializedReport.entries,
+                    outputs: serializedReport.outputs,
+                });
+                writeFile('logs', serializedReport.logs);
+                writeFile('timings', serializedReport.timings);
+                writeFile('dependencies', serializedReport.inputs);
```
jakub-g, line 118: just to make sure, no typo here?

packages/plugins/output/src/validate.test.ts
```diff
@@ -0,0 +1,117 @@
...
+                expected: true,
+            },
+            {
+                description: 'return false when output config has enable: false',
+                input: { output: { enable: false } },
+                expected: false,
+            },
+        ];
+
+        test.each(cases)('Should $description', ({ input, expected }) => {
```
jakub-g, line 32: Is `$description` some magic feature of test runner?

jakub-g: I never saw this; googling around I found some [blogs](https://www.jacobparis.com/content/dynamically-generate-test-cases) and StackOverflows telling that it should be `%` not `$`

packages/plugins/output/src/index.ts
yoannmoinet: No typo, all good, it just happens that the `inputs` list of the build report also acts as the dependency graph of the project.

packages/plugins/output/src/validate.test.ts
yoannmoinet: From the [jest docs](https://jestjs.io/docs/api#1-describeeachtablename-fn-timeout)

> Or generate unique test titles by injecting properties of test case object with $variable
To inject nested object values use you can supply a keyPath i.e. $variable.path.to.value (only works for "own" properties, e.g. $variable.constructor.name wouldn't work)
You can use $# to inject the index of the test case
You cannot use $variable with the printf formatting except for %%

PR #206, opened 2025-08-12
[chore] v3 merge

yoannmoinet:


yoannmoinet:
Did a rebase instead, worked better.

PR #208, opened 2025-08-12
[chore] Create self.bump.sts.yaml

yoannmoinet:
Need this to use dd-octo-sts in https://github.com/DataDog/build-plugins/pull/207


Ayc0 (approved):
image

PR #224, opened 2025-09-25
Upgrade js-instrumentation lib version

cy-moi:
### What and why?
We need to include the new updates from @datadog/js-instrumentation-wasm.

### How?
Update the version of @datadog/js-instrumentation-wasm in dependencies.


yoannmoinet (approved):
lgtm

PR #227, opened 2025-10-01
[fix] Gate banner and footer injections to only supported file types for esbuild

yoannmoinet:
### What and why?

The injection plugin would crash if it tried to inject content into CSS files.

### How?

Gate the banner/footer injections to only supported files.

### DIFF
--- a/packages/plugins/injection/src/esbuild.ts
+++ b/packages/plugins/injection/src/esbuild.ts
@@ -119,16 +119,15 @@ export const getEsbuildPlugin = (
                 })
                 .filter(Boolean) as string[];
 
-            const isSupported = (file: string): boolean => {
-                const { ext } = path.parse(file);
+            const isSupported = (ext: string): boolean => {
                 return ['.js', '.ts', '.tsx', '.jsx'].includes(ext);
             };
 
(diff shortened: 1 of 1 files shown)


packages/plugins/injection/src/esbuild.ts
```diff
@@ -117,28 +119,42 @@ export const getEsbuildPlugin = (
...
-                    if (isNodeSystemError(e) && e.code === 'ENOENT') {
-                        // When we are using sub-builds, the entry file of sub-builds may not exist
-                        // Hence we should skip the file injection in this case.
-                        log.warn(`Could not inject content in ${output}: ${e}`);
-                    } else {
-                        throw e;
+            const proms = outputs
+                .filter((output) => {
+                    const isOutputSupported = isSupported(output);
+                    const { base, ext } = path.parse(output);
```
sethfowler-datadog, line 131: Nit: It might be nice to pass the return value of `path.parse()` to `isSupported()`, instead of the original filename; this way you wouldn't have to parse the path twice in succession.

sethfowler-datadog (approved):
LGTM!

PR #233, opened 2025-10-10
[fix] Patch privacy plugin with two minor changes

cy-moi:
### What and why?
We noticed some issue while using the privacy plugin in web-ui build:
- If the dictionary is empty, we do not extract the allowlist or transform the code. Hence, we could use the original code here to make the behavior explicit.
- We did not add the dictionary size in log message context either and this prevented us to monitor it effectively.
- Handling jsx attribute line breaker in `js-instrumentation-wasm` v1.0.8

### How?
1. Return original code in transform hook if dictionary size is 0
2. Add dictionary size in build end log message attributes (new logs look like [this](https://app.datadoghq.com/logs?query=%40dictionarySize%3A%2A&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&event=AwAAAZnN53vwd20zmAAAABhBWm5ONTN2d0FBQmJZZTB1YzlEUEhBQUEAAAAkZjE5OWNkZTctYjAzNi00MGU4LWJhOTYtYWVhNDljMTZlNTY3AAAAAw&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1760094603646&to_ts=1760095503646&live=true))
3. Upgrade `js-instrumentation-wasm` version

### DIFF
--- a/packages/plugins/rum/src/privacy/index.ts
+++ b/packages/plugins/rum/src/privacy/index.ts
@@ -19,7 +19,8 @@ export const getPrivacyPlugin = (
         pluginOptions.helperCodeExpression,
         context.bundler.name,
     );
-    let dictionarySize = 0;
+    let dictionaryEntryCount = 0;
+    let fileCount = 0;
     return {
         name: PLUGIN_NAME,
         // Enforce when the plugin will be executed.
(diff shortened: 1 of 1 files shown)


packages/plugins/rum/src/privacy/index.ts
```diff
@@ -49,6 +55,9 @@ export const getPrivacyPlugin = (
         buildEnd: () => {
             log.debug(`Instrumentation result - privacy dictionary size: ${dictionarySize}`, {
```
sethfowler-datadog: This is a bit misleading; we are adding up the size of the privacy dictionary in each individual file and calling it the total privacy dictionary size, but in fact, duplicate entries in different files are deduplicated at runtime, so the actual privacy dictionary size will be smaller.

It might make sense to reword this to better clarify what's happening; perhaps something like:
```suggestion
            log.debug(`Privacy dictionary will include ${dictionaryEntryCount} entries across ${fileCount} files`, {
```

sethfowler-datadog (approved):
LGTM!

PR #296, opened 2026-04-01
Add live-debugger plugin

watson:
Supersedes #253

### What and why?

Adds a new **Live Debugger** build plugin (`@dd/live-debugger-plugin`) that automatically instruments JavaScript/TypeScript functions at build time to enable Live Debugger without requiring code rebuilds.

When enabled, every matching function in the application is wrapped with lightweight probes that can be activated at runtime via `$dd_probes(functionId)`. When no probes are active the check returns `undefined` and all instrumentation is skipped.

This originated as an [Innovation Week POC](https://github.com/DataDog/build-plugins/pull/253) and has been hardened with bug fixes, performance optimizations (down from ~17x to ~1.2x build overhead), comprehensive test coverage, and filtering options.

**Notes:**
- This PR is not intended to be production-ready. The goal is to land the plugin in a state that is ready for dogfooding. In particular, the function ID algorithm (`<relative-file-path>;<function-name>`) is a placeholder — the final stable ID scheme will be implemented in a follow-up PR before this is ready for end user consumption.
- There is also a known escaping limitation in the current POC: function IDs are still embedded directly into generated code, so quoted method names containing a single quote can produce invalid output. This will be fixed before the plugin is considered production-ready.

### Overhead stats

- Build overhead: ~1.2x (for the files where instrumentation code is added).
- Runtime overhead of no-op instrumented functions is negligible (near zero) for all browsers except Firefox, which isn't able to fully optimize away the no-op instrumentations.

### Installation

Babel (`@babel/parser`, `@babel/traverse`, `@babel/types`) and `magic-string` are declared as **optional peer dependencies** on the 5 published bundler packages. Users who don't enable Live Debugger don't pay the install-size cost (~12 MB of Babel + `magic-string`). When the plugin is enabled (by providing a `liveDebugger` configuration), projects install the peer deps alongside the Datadog plugin:

```bash
npm install --save-dev @babel/parser @babel/traverse @babel/types magic-string
```

If any of them is missing at build time, the plugin throws with an actionable install hint instead of a raw Node `MODULE_NOT_FOUND`.

### How?

**New plugin (`packages/plugins/live-debugger/`)**

- **Transform pipeline** (`src/transform/`): Uses Babel to parse the AST in read-only mode, collects instrumentation targets, then applies injections via MagicString (no AST mutation). Processes inner functions before outer ones so that `appendLeft` calls at shared positions stack correctly.
- **Lazy loader** (`src/transform/loader.ts`): Gates the entire transform module behind `getTransformCode()`, so Babel and `magic-string` are not loaded at all when the `liveDebugger` configuration is omitted. When the plugin is enabled, peer deps are loaded on the first file that reaches the transform hook, and a missing module is rewrapped with a clear install hint.
- **Function ID generation** (`src/transform/functionId.ts`): Produces human-readable IDs in `<relative-file-path>;<function-name>` format. Anonymous functions use `<anonymous>@line:col:siblingIndex` for disambiguation. This algorithm is temporary and will be replaced with a stable, production-grade scheme in a follow-up.
- **Scope tracking** (`src/transform/scopeTracker.ts`): Extracts variable names (params + locals, capped at 25) for entry and exit snapshots so probes can capture local state.
- **Instrumentation guards** (`src/transform/instrumentation.ts`): Skips functions that can't be safely instrumented (generators, async generators, class constructors) and respects `// @dd-no-instrumentation` skip comments.
- **CJS interop** (`src/transform/cjs-interop.ts`): Handles `@babel/traverse` CJS default-export resolution in bundled environments where the module wrapper shape can vary.
- **Option validation** (`src/validate.ts`): Validates all user-supplied config with descriptive error messages.

**Configuration options** (providing a `liveDebugger` object — even an empty `{}` — enables the plugin, matching the convention used by the other product plugins):
- `include` / `exclude` — file patterns to control scope (defaults include `.js/.jsx/.ts/.tsx`, exclude `node_modules`, minified files, virtual modules, Datadog SDK packages, etc.)
- `honorSkipComments` — respect `// @dd-no-instrumentation` comments (default: `true`)
- `functionTypes` — restrict to specific function kinds (e.g., `functionDeclaration`, `arrowFunction`, `classMethod`)
- `namedOnly` — skip anonymous functions (default: `false`)

The env variable `DD_LD_LIMIT` can cap the number of files with functions that are processed, as a safety valve during dogfooding.

**Factory integration:**
- Registered in `packages/factory/src/index.ts` and `packages/core/src/types.ts` via the standard plugin injection markers.
- Exposed through all published bundler packages (webpack, esbuild, rollup, rspack, vite), with Babel and `magic-string` declared as optional `peerDependencies` so they're only installed by consumers who enable Live Debugger.
- Hidden from the root README via `hideFromRootReadme: true` until the plugin is production-ready.

**Runtime stubs:**
- The plugin injects a minimal no-op stub (~80 bytes) into all output chunks via `context.inject()`. This defines `globalThis.$dd_probes` as an empty function so instrumented code never crashes when the Datadog Browser Debugger SDK (`@datadog/browser-debugger`) is absent.
- `$dd_entry`, `$dd_return`, and `$dd_throw` don't need stubs because they are always guarded by `if (probe)` checks in the injected instrumentation.
- When the SDK loads and `DD_DEBUGGER.init()` is called, it overwrites `$dd_probes` with the real implementation and sets up the other three globals. Probes activate immediately — no rebuild required.

**Testing & benchmarks:**
- Comprehensive transform unit tests covering all supported function types, instrumentation patterns, skip comments, control flow, source maps, and deterministic output
- Plugin unit tests covering child-compilation include/exclude fallback behavior, `DD_LD_LIMIT` capping, error handling, and option validation
- CJS interop tests and a dedicated peer-dependency test suite (eager-import surface, lazy-load chain, missing-dep diagnostics) that locks in the optional-install contract
- E2E browser smoke tests running across all supported bundlers and browsers, verifying the app works both without the debugger SDK and with simulated active probes (including code-splitting with dynamic imports)
- Benchmark script (`scripts/benchmark-subset.js`) for measuring build overhead on real file subsets


watson:
* **#296** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/296?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#324](https://github.com/DataDog/build-plugins/pull/324) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/324?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#325](https://github.com/DataDog/build-plugins/pull/325) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/325?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/296?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/published/vite-plugin/package.json
```diff
@@ -50,12 +50,16 @@
         "watch": "yarn build --watch"
     },
     "dependencies": {
+        "@babel/parser": "^7.23.0",
+        "@babel/traverse": "^7.23.0",
+        "@babel/types": "^7.23.0",
```
yoannmoinet: Can you use fixed version please.
Also, you can use versions that we already have in (through the `@babel/core` and others in devDependencies.


```suggestion
        "@babel/parser": "7.24.5",
        "@babel/traverse": "7.24.5",
        "@babel/types": "7.24.5",
```

packages/published/vite-plugin/package.json
```diff
@@ -50,12 +50,16 @@
...
         "@datadog/js-instrumentation-wasm": "1.0.8",
         "async-retry": "1.3.3",
         "chalk": "2.3.1",
         "glob": "11.1.0",
         "json-stream-stringify": "3.1.6",
         "jszip": "3.10.1",
+        "magic-string": "0.30.21",
```
yoannmoinet: We have 0.30.3 in already.
So the footprint of the PR is smaller.
(Or update everything to `0.30.21` 😄)

```suggestion
        "magic-string": "0.30.3",
```

yoannmoinet (commented):
Pretty solid PR, well done!

I have some concerns though.

A ~1.2x build overhead is not "negligible", that's 20%.
Totally not a blocker, but I don't want to misrepresent the impact of this.

I'm not a fan of adding babel for everyone, especially with Unplugin shipping with [rollup's `this.parse`](https://rollupjs.org/plugin-development/#this-parse) for both rollup and vite (and rolldown), which makes babel redundant in these cases.

It could also be set as peer dependencies or optional dependencies, and verified at runtime. So not everyone installing `build-plugins` has to pay the price of such dependencies.

[ref](https://unplugin.unjs.io/guide/#supported-context)
> For bundlers other than Rollup, Rolldown, or Vite, `setParseImpl` must be called to manually provide a parser implementation. Parsers such as [Acorn](https://github.com/acornjs/acorn), [Babel](https://babeljs.io/), or [Oxc](https://oxc.rs/) can be used.

I'd prefer you to follow this way of adding a parser to the ecosystem, could sit in a different, independent PR too, it would lighten this one.

Note that we have a parallel effort in that regard, led by @sdkennedy2, and they want to use oxc, which is more modern than babel imo, and a better option. For instance, oxc would prevent some workarounds you have to do with babel.

watson:
> A ~1.2x build overhead is not "negligible", that's 20%.
> Totally not a blocker, but I don't want to misrepresent the impact of this.

Sorry, I was referring to the runtime overhead being negligible. But I can see the PR body could easily be misunderstood. I've re-phrased it.

packages/published/vite-plugin/package.json
watson: Good point. I'll bump the version in `yarn.lock` for everyone.

packages/published/vite-plugin/package.json
watson: All the other `package.json` files uses `^7.23.0` - that's why I did that as well here. I'll make sure it resolves to the same versions in `yarn.lock` though

yoannmoinet: > All the other package.json files uses ^7.23.0 

Which ones? I only see it in the ones modified in this PR.

Sorry I didn't elaborate more, I only commented here, but it does apply to all the `package.json` changed in this PR 😅

watson: Sorry, my bad. I got confused and thought for a moment that the changes to `packages/published/*/package.json` was pre-existing. But I see now that they are not. I'll fix it.

packages/published/vite-plugin/package.json
watson: Fixed 👍

packages/published/vite-plugin/package.json
watson: Fixed 👍

packages/plugins/live-debugger/src/validate.ts
```diff
@@ -0,0 +1,100 @@
...
+
+    // Throw if there are any errors
+    if (errors.length) {
+        log.error(`\n  - ${errors.join('\n  - ')}`);
+        throw new Error(`Invalid configuration for ${PLUGIN_NAME}.`);
+    }
+
+    // Build the final configuration with defaults
+    return {
+        enable: !!pluginConfig.enable,
```
yoannmoinet: This is kind of different from all the other plugins that default to `true`.
Basically, using an empty object will enable the plugin.
Can you follow the same logic here please?

packages/plugins/live-debugger/README.md
```diff
@@ -0,0 +1,225 @@
...
+        const $dd_rv = x * 2;
+        if ($dd_p) $dd_return($dd_p, $dd_rv, this, $dd_e(), $dd_e());
+        return $dd_rv;
+    } catch(e) { if ($dd_p) $dd_throw($dd_p, e, this, $dd_e()); throw e; }
+};
+```
+
+### liveDebugger.enable
+
+> default: `false`
```
yoannmoinet: This should be `true` by default, to follow the same pattern as other plugins.

```suggestion
> default: `true`
```

yoannmoinet (approved):
Just a nit on the pattern we use for `enable` in other plugins.
Other than that, lgtm.

Especially if it's not "public" yet, and that it uses peer/optional dependencies for `@babel/*`.

I won't be as strict. Outside's impact is pretty minimal.

packages/plugins/live-debugger/src/validate.ts
watson: Fixed 👍

packages/plugins/live-debugger/README.md
```diff
@@ -116,12 +115,6 @@ const double = (x) => {
 };
 ```
 
-### liveDebugger.enable
-
-> default: `false`
-
-Enable or disable Live Debugger. When enabled, all matching JavaScript files will be instrumented at build time.
-
```
yoannmoinet: You should keep this in, just changing the default to `true`.

packages/plugins/live-debugger/src/index.test.ts
```diff
@@ -294,15 +294,6 @@ describe('getLiveDebuggerPlugin', () => {
...
 describe('getPlugins', () => {
-    it('should return an empty array when enable is false', () => {
-        const arg = getGetPluginsArg({ liveDebugger: { enable: false } });
-
-        const plugins = getPlugins(arg);
-
-        expect(plugins).toEqual([]);
-        expect(arg.context.inject).not.toHaveBeenCalled();
-    });
-
```
yoannmoinet: This is still valuable to have I think.

packages/plugins/live-debugger/README.md
watson: Oh, I accidentally looked at the docs for error-tracking where this property wasn't documented. I thought it was on purpose. I've re-added it and will make a separate PR to align the docs across the repo related to this.

packages/plugins/live-debugger/src/index.test.ts
watson: Fixed 👍

packages/plugins/live-debugger/src/validate.ts
```diff
@@ -15,6 +15,11 @@ export const validateOptions = (config: Options, log: Logger): LiveDebuggerOptio
     const pluginConfig: LiveDebuggerOptions = config[CONFIG_KEY] || {};
     const errors: string[] = [];
 
+    // Validate enable option
+    if (pluginConfig.enable !== undefined && typeof pluginConfig.enable !== 'boolean') {
+        errors.push(`${red('enable')} must be a boolean`);
+    }
```
yoannmoinet, line 21: A bit overkill but that's fine haha.

yoannmoinet (approved):
lgtm. :shipit:

PR #317, opened 2026-04-07
[APPS] Use manualChunks for self-contained backend function builds

sdkennedy2:
## Motivation

PR #317 originally moved `inlineDynamicImports: true` into the shared backend build config and refactored production builds to loop one `vite.build()` per function. A coworker pointed out that `manualChunks: () => '[name]'` — a pattern already used in this repo's own rollup config — achieves the same "no code splitting" goal while supporting multiple entry points in a single build call.

## Changes

Replaced `inlineDynamicImports: true` with `manualChunks: () => '[name]'` in the shared backend build config. This lets us batch all backend functions into a single `vite.build()` call instead of looping one call per function.

The key difference: `inlineDynamicImports` only works with a single entry point (Rollup limitation), forcing us to build each function individually. `manualChunks` assigns all modules to the entry chunk, preventing code splitting while supporting multiple entries. This is the same pattern used at `packages/tools/src/rollupConfig.mjs:193`.

**`build-config.ts`**: Swapped `inlineDynamicImports: true` for `manualChunks: () => '[name]'` and moved `entryFileNames: '[name].js'` into the base config.

**`build-backend-functions.ts`**: Replaced the per-function loop (N `vite.build()` calls) with a single batched build — all virtual entries and inputs are collected up front, one build produces all outputs. Removes the `no-await-in-loop` eslint disable, `emptyOutDir: false` workaround, and per-function overrides.

**`dev-server.ts`**: No changes — builds one function per request and inherits the new config automatically.

## QA Instructions

1. Scaffold or use an existing test app with 2+ backend functions
2. Link local build-plugins via `yarn dev`
3. Run `vite build` — verify each backend function produces a self-contained `.js` file
4. Run `vite` dev server — verify `/__dd/executeAction` still works for each function

## Blast Radius

- Only affects the apps plugin's backend function build pipeline
- No changes to the main client build or any other plugin
- The dev server path is unchanged (still builds one function per request)

## Documentation

### DIFF
--- a/packages/plugins/apps/src/vite/build-backend-functions.ts
+++ b/packages/plugins/apps/src/vite/build-backend-functions.ts
@@ -28,55 +28,46 @@ export async function buildBackendFunctions(
 ): Promise<string> {
     const outDir = await mkdtemp(path.join(tmpdir(), 'dd-apps-backend-'));
 
-    const virtualEntries: Record<string, string> = {};
-    const input: Record<string, string> = {};
-
-    for (const func of functions) {
-        const virtualId = `${VIRTUAL_PREFIX}${func.name}`;
-        virtualEntries[virtualId] = generateVirtualEntryContent(
-            func.name,
-            func.entryPath,
-            buildRoot,
-        );
-        input[func.name] = virtualId;
-    }
-
     log.debug(`Building ${functions.length} backend function(s) via vite.build()`);
 
-    const baseConfig = getBaseBackendBuildConfig(buildRoot, virtualEntries);
+    // Build each function individually so that each output is a single
+    // self-contained JS file
+    for (const func of functions) {
+        const virtualId = `${VIRTUAL_PREFIX}${func.name}`;
+        const virtualContent = generateVirtualEntryContent(func.name, func.entryPath, buildRoot);
 
-    // Production: build all functions in one vite.build() call, writing each to
-    // disk as a named file so the archive/upload step can collect them.
-    // Uses multi-entry input (one per function) with \0-prefixed virtual IDs —
-    // the \0 convention prevents other plugins from processing these IDs.
-    const result = await viteBuild({
-        ...baseConfig,
-        build: {
-            ...baseConfig.build,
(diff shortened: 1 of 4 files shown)


sdkennedy2:
* **#317** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/317?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/317?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#291** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/291?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#306](https://github.com/DataDog/build-plugins/pull/306) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/306?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#292** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/292?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

yoannmoinet (approved):
Since we are working on customer's machine with this, I'd advocate for tight performances on this.

Not needed right away in this PR directly (so we can ship it asap), but in a soon-ish follow-up to have the builds made in parallel instead of sequentially, so its faster for the user.

Since this is something susceptible to run in CI, it will need some queueing system, in order not to flood the CPUs and memory capacities.

Multiple leads for it:

- the internal `async-queue` plugin, that allows you to defer the build and let the lifecycles of the build-plugins continue uninterrupted, but still hold the build process until these are finished. Works as a fire-and-forget deferral.
- `PQueue` usage with `concurrency`, [example](https://github.com/DataDog/build-plugins/blob/2b3b6ab6fdc9f124b5b099a5b7eb51effb79b83e/packages/plugins/error-tracking/src/sourcemaps/sender.ts#L89)

PR #320, opened 2026-04-14
[APPS] Support importing .backend.ts files as RPC proxies from frontend code

sdkennedy2:
## Motivation

Backend RPC in `@dd/apps-plugin` was tightly coupled to a fixed `backend/` directory, one-function-per-file assumptions, and a separate runtime package path.

This PR shifts backend function discovery to the bundler transform pipeline so functions can be declared with a `.backend.{ts,tsx,js,jsx}` extension anywhere in the project, and the plugin generates frontend call proxies automatically. It also moves backend execution runtime ownership into the apps plugin.

### Transform-based backend discovery
- Added a transform hook for `.backend.ts|.tsx|.js|.jsx` files.
- Replaced pre-scan discovery with `this.parse(code)` AST-based export extraction.
- Registered discovered backend functions in a registry keyed by entry path for HMR-safe updates.
- Replaced matched backend modules with generated proxy code during transformation.

### Proxy generation and query-name encoding
- Generated proxy functions for every named export, e.g.:
  - `export async function <name>(...args)` → `globalThis.DD_APPS_RUNTIME.executeBackendFunction("<queryName>", args)`.
- Used opaque query IDs from `encodeQueryName` (`sha256(relativePath).exportName`) so frontend bundles do not leak backend source paths.

### Runtime injection and execution transport
- Added a built runtime module that sets `globalThis.DD_APPS_RUNTIME.executeBackendFunction` in the page, so generated proxy modules call a single shared runtime API instead of importing a package.
- `executeBackendFunction` dispatches calls based on context:
  - In App Builder iframe: sends `postMessage` to parent and waits for query response.
  - In local development: calls the local endpoint `POST /__dd/executeAction`.
- Runtime injection uses the same pattern already used by the RUM plugin for app initialization.
- Switched ownership of runtime logic into the apps plugin, removing dependency on `@datadog/apps-function-query`.

### Build and packaging integration
- Backend bundle and dev execute flows now share the same backend-function metadata model.
- Backend bundles are emitted under `backend/<queryName>.js`; frontend assets stay under `frontend/`.
- Archive path handling now normalizes to forward slashes for cross-platform package correctness.

## QA Instructions

1. Run unit tests:
   - `yarn workspace @dd/tests test:unit packages/plugins/apps`
2. Validate dev execution:
   - Open a sample app, trigger a backend proxy call from frontend.
   - Verify `window.DD_APPS_RUNTIME.executeBackendFunction` exists.
   - Verify `POST /__dd/executeAction` returns success in dev flow.
3. Verify generated proxy output does not include raw backend file paths.

## Blast Radius

- Backend files now use `.backend.ts|.tsx|.js|.jsx` and are no longer required under a fixed `backend/` directory.
- Multiple named exports per backend file are now supported.
- Existing `backendDir` behavior/config is removed.
- `@datadog/apps-function-query` is no longer required by the apps/vite plugin.
- Apps now rely on `globalThis.DD_APPS_RUNTIME` as the runtime entrypoint for backend execution calls.

## Documentation

- No doc changes included yet; recommend adding a short migration note in plugin docs:
  - moving from `backendDir`/fixed path model to `.backend.*`
  - new naming and multi-export behavior
  - sample before/after of a backend function and generated proxy usage.


sdkennedy2:
* **#329** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/329?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#323** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/323?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#320** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/320?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/320?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#322** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/322?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/package.json
```diff
@@ -29,6 +29,7 @@
         "pretty-bytes": "5.6.0"
     },
     "devDependencies": {
+        "@types/estree": "1.0.8",
```
sdkennedy2, line 32: It seems that unplugin doesn't export the estree types that this.parse provides. I had to add this as a dev dep. I couldn't find a way around adding this but let me know if you have thoughts.

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/backend/discovery.ts
```diff
@@ -3,64 +3,170 @@
...
+
+export interface BackendFunction {
+    /** The BackendFunctionRef identifying this function */
+    ref: BackendFunctionRef;
+    /** Absolute path to the .backend.ts source file */
     entryPath: string;
 }
 
-const EXTENSIONS = ['.ts', '.js', '.tsx', '.jsx'];
+export interface BackendFileInfo {
```
sdkennedy2: I'm going to combine these into one type

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/backend/discovery.ts
```diff
@@ -2,65 +2,112 @@
...
+    const names: string[] = [];
+    for (const node of ast.body) {
+        // handles: export default ...
+        if (node.type === 'ExportDefaultDeclaration') {
+            throw new Error(
+                `Default exports are not supported in .backend.ts files. Use a named export instead: ${filePath}`,
+            );
+        }
+        if (node.type !== 'ExportNamedDeclaration') {
+            continue;
```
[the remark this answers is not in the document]
yoannmoinet: @sdkennedy2 Maybe it's better to throw here for now, and not support this pattern? WDYT?

packages/plugins/apps/src/backend/discovery.ts
```diff
@@ -2,65 +2,121 @@
...
-                const name = path.basename(entry, ext);
-                functions.push({ name, entryPath });
+            // handles: export { add as default }
+            if (spec.exported.name === 'default') {
+                throw new Error(
+                    `Default exports are not supported in .backend.ts files. Use a named export instead: ${filePath}`,
+                );
             }
+            // handles: export { add, multiply }
+            names.push(spec.exported.name);
```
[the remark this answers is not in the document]
yoannmoinet: @sdkennedy2 sounds like a legit concern, isn't it?

packages/plugins/apps/src/vite/index.ts
```diff
@@ -33,14 +33,15 @@ export interface VitePluginOptions {
...
     backendOutputs,
     handleUpload,
     log,
     auth,
 }: VitePluginOptions): PluginOptions['vite'] => ({
     async closeBundle() {
         let backendOutDir: string | undefined;
+        const functions = getBackendFunctions();
         if (functions.length > 0) {
             backendOutDir = await buildBackendFunctions(
```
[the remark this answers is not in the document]
yoannmoinet: Given that we specifically support the dev mode, this looks legit too.

packages/plugins/apps/src/vite/proxy-codegen.ts
```diff
@@ -0,0 +1,38 @@
...
+ * code — only the hashed query name appears, preventing backend file
+ * structure from leaking into frontend bundles.
+ *
+ * @param exports - The export name + pre-computed query name for each export
+ * @returns Generated proxy module source code
+ */
+export function generateProxyModule(exports: ProxyExport[]): string {
+    const lines: string[] = [];
+
+    lines.push("import { executeBackendFunction } from '@datadog/apps-function-query';");
```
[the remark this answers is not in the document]
yoannmoinet: Doesn't this mean we need to add `@datadog/apps-function-query` in all published's `package.json`? 
(Well, at least `@datadog/vite-plugin` since we only support this one for now)

packages/plugins/apps/src/backend/discovery.ts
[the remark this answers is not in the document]
yoannmoinet: It's most likely the same issue for `export class MyClass...`.

packages/plugins/apps/src/index.ts
```diff
@@ -148,23 +188,51 @@ Either:
...
+                // For each .backend.* file, parse its named exports, register
+                // them as backend functions, and replace the module with a
+                // frontend proxy that calls executeBackendFunction at runtime.
+                handler(code, id) {
+                    const exportNames = extractExportedFunctions(this.parse(code), id);
+                    if (exportNames.length === 0) {
+                        // Clear any previously registered functions for this file
+                        // so stale entries don't persist across HMR re-transforms.
+                        setBackendFunctions(id, []);
+                        return undefined;
```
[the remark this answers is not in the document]
yoannmoinet: There seems to be a gap here (ref codex's review).

packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -349,6 +363,8 @@ export function createDevServerMiddleware(
             return;
         }
 
+        const functionsByName = buildFunctionMap(getBackendFunctions());
```
yoannmoinet, line 371: Is the performance hit here acceptable?
(Rebuilding the map at every request)

packages/plugins/apps/src/index.ts
```diff
@@ -148,23 +188,51 @@ Either:
...
+                        include: [BACKEND_FILE_RE],
+                        exclude: [/node_modules/, /[/\\]dist[/\\]/],
+                    },
+                },
+                // For each .backend.* file, parse its named exports, register
+                // them as backend functions, and replace the module with a
+                // frontend proxy that calls executeBackendFunction at runtime.
+                handler(code, id) {
+                    const exportNames = extractExportedFunctions(this.parse(code), id);
+                    if (exportNames.length === 0) {
```
yoannmoinet, line 210: I feel like this could warrant for a warning.

Having a `*.backend.ts` file exporting nothing could be a mistake from the user, better make them aware of it.

packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -308,27 +317,32 @@ async function handleExecuteAction(
...
-
     const bundle = (func: BackendFunction, args: unknown[]) =>
         bundleBackendFunction(viteBuild, func, args, projectRoot, log);
 
-    if (backendFunctions.length > 0) {
-        log.info(
-            `Dev server middleware active for ${backendFunctions.length} backend function(s): ${backendFunctions.map((f) => f.name).join(', ')}`,
-        );
-    }
+    log.info('Dev server middleware active for backend functions');
```
yoannmoinet: Not sure why you changed this?
This feels unnecessary noisy if you don't have backend functions.

packages/plugins/apps/src/backend/discovery.ts
```diff
@@ -2,65 +2,121 @@
...
+import type { Declaration, Expression, Program } from 'estree';
+import type { AstNode } from 'rollup';
 
 export interface BackendFunction {
+    /** Relative path from project root to the .backend.ts file (without extension) */
+    path: string;
+    /** Exported function name */
     name: string;
+    /** Absolute path to the .backend.ts source file */
     entryPath: string;
```
yoannmoinet: I'd be more explicit in the naming of both `path` attributes.

```suggestion
    relativePath: string;
    /** Exported function name */
    name: string;
    /** Absolute path to the .backend.ts source file */
    absolutePath: string;
```

yoannmoinet (commented):
Looks good overall, I just have some comments.

I'd also address Codex's review.

I really like the `*.backend.ts` approach.

packages/plugins/apps/src/backend/discovery.ts
sdkennedy2: Yeah I had a similar thought. I added the JSDoc comments to help but I agree more explicit names would be better 👍.

packages/plugins/apps/src/backend/discovery.ts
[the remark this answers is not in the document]
sdkennedy2: I'd be down to throw. This seems like a good idea.

packages/plugins/apps/src/backend/discovery.ts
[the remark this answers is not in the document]
sdkennedy2: Yes I'd agree.

packages/plugins/apps/src/vite/dev-server.ts
sdkennedy2: Good call

packages/plugins/apps/src/vite/dev-server.ts
sdkennedy2: I don't think it would be too bad since all we are doing is the following. Is it the hashing of the path that you are concerned about?
```
const functionsByName = new Map(backendFunctions.map((f) => [encodeQueryName(f), f]))
```

yoannmoinet: It's mostly because the behaviour changed.
Before we were building the map only on server start.
Now it's done on each request, which is significantly more times.

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/backend/discovery.ts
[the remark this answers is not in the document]
sdkennedy2: This should be fixed now because we throw when we receieve `export * from './impl'`

packages/plugins/apps/src/backend/discovery.ts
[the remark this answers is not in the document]
sdkennedy2: We name handle local declarations so that if you have:
```typescript
const VERSION = '1.0'; export { VERSION }; 
```
It will correctly check if the VERSION variable declaration is callable. In this case we will throw.

The only area thats a little shaky is we allow the following but we don't verify that handler is callable. It seemed like a reasonable tradeoff though. Alternatively we could throw in this scenario.
```typescript
import {handler} from '..';
export {handler};
```

packages/plugins/apps/src/vite/dev-server.ts
sdkennedy2: I think it should be fine. Even if the number of backend functions is ~1000 iterating over an array that size, generating a map and hashing a path that wouldn't be more than ~200 character doesn't seem that bad. It seems like this could be accomplished in the order of microseconds.

packages/plugins/apps/src/vite/proxy-codegen.ts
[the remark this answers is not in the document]
sdkennedy2: I added `@datadog/apps-function-query` based on our conversation, but I think I'll want to bump the version of `@datadog/apps-function-query` to `0.0.2` and update this PR before landing it.

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

packages/tools/src/commands/integrity/dependencies.ts
```diff
@@ -31,6 +31,19 @@ const dependencyExceptions = [
     '@dd/tools',
 ];
 
+// Dependencies that are added directly to a published package and should not
+// be removed by the integrity check, even though they don't come from an
+// internal plugin's dependency tree.
+const publishedPackageExtraDependencies: Record<string, Record<string, string>> = {
```
sdkennedy2, line 37: Let me know if this is an acceptable addition for now.

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/index.ts
```diff
@@ -148,23 +205,54 @@ Either:
...
-            handleUpload,
-            log,
-            auth: context.auth,
-        }),
-    });
+    return [
+        {
+            name: PLUGIN_NAME,
+            enforce: 'post',
+            transform: {
```
yoannmoinet, line 219: 🤔 isn't this executed for all bundlers, and not just vite?
How impactful is this hook on the bundled code?

sdkennedy2: You're right. The transform is at the top of the plugin object, not under vite:, so it fires for non-vite bundlers too. In those bundlers the .backend.* files get rewritten into proxies that call DD_APPS_RUNTIME.executeBackendFunction, but getVitePlugin's closeBundle is the only thing that actually builds and uploads the backend bundles. This means the client ships working-looking proxies pointing at a backend that was never produced. Fails silently at runtime rather than at build time.

Thoughts on early returning [] within getPlugins when context.bundler.name !== 'vite' with a log.warn so users configuring apps under a non-vite bundler get a clear "apps is vite-only" signal up front?

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/index.ts
```diff
@@ -36,13 +84,29 @@ export const getPlugins: GetPlugins = ({ options, context, bundler }) => {
         return [];
     }
 
-    // Discover backend functions (sync — must run before build starts).
-    const absoluteBackendDir = path.resolve(context.buildRoot, validatedOptions.backendDir);
-    const backendFunctions = discoverBackendFunctions(absoluteBackendDir, log);
-    const backendOutputs = new Map<string, string>();
-    const hasBackend = backendFunctions.length > 0;
+    if (context.bundler.name !== 'vite') {
```
sdkennedy2, line 87: Let me know your thoughts on this. Alternatively I could move the transform into getVitePlugin, but it seems the current tranform code could work for other bundlers once we build out the remaining missing functionality.

yoannmoinet (approved):

PR #324, opened 2026-04-19
Centralize `enable` option handling in the factory

watson:
### What and why?

The six user-facing plugins (`apps`, `error-tracking`, `live-debugger`, `metrics`, `output`, `rum`) each had a divergent implementation for resolving their `enable` config boolean:

- **Two code patterns:** `apps` and `live-debugger` used explicit `??` nullish coalescing, while the other four used a spread-override trick (`enable: !!config[KEY], ...config[KEY]`) that relies on property ordering.
- **Inconsistent validation:** Only `live-debugger` rejected non-boolean values; the rest silently coerced.
- **Incomplete docs:** `output` and `metrics` documented `default: true` (misleading — it's false when the config block is absent), and `error-tracking` had no `enable` section at all.

This PR makes the factory the single owner of `enable` resolution so every plugin shares one set of semantics — and individual plugins stop having to think about `enable` at all.

### How?

**Factory becomes the gate (`@dd/factory`):**
Each user-facing plugin's `<configKey>.enable` is resolved once in the factory's plugin loop, before the plugin's `getPlugins` is even called. If it resolves to `false`, the plugin is never instantiated, never validated, never hooked in. As far as the rest of the build is concerned, it doesn't exist.

**Shared helper (`@dd/core/helpers/options`):**
- `resolveEnable(options, configKey, log)` — single source of truth used by the factory. Emits a deprecation warning (once per key) when a non-boolean value is passed, coercing it with `!!` to avoid breaking existing users. A `TODO(next major)` marks the coercion site for removal.

**Plugins simplified:**
- `getPlugins` no longer has an `if (!validatedOptions.enable) return []` guard.
- `validateOptions` no longer calls `resolveEnable`; `apps`, `output`, and `metrics` no longer need a `log` parameter at all.
- Each `<Plugin>OptionsWithDefaults` type drops `enable` — the factory guarantees it's `true` when plugin code runs, so it's meaningless in the resolved shape.
- `live-debugger` now uses the same soft coercion as everyone else (with a deprecation warning) instead of throwing on non-boolean values. Strict rejection moves to the next major along with the others.

**Integrity tooling:**
The `#configs-injection-marker` block now emits `[name, CONFIG_KEY, getPlugins]` triples so the factory's enable loop stays in sync as plugins are added or removed. The `create-plugin` scaffolding follows the same pattern (no `enable` guard, no `enable` in `OptionsWithDefaults`).

**Documentation:**
- Every user-facing plugin's README now has an identical `### <key>.enable` section with the correct default wording: ``true`` when a `<key>` config block is present, ``false`` otherwise.
- `error-tracking` gained its missing `enable` section and TOC entry.
- `live-debugger`'s README updated to reflect the soft coercion (was "Must be a boolean").

**Tests:**
- Factory gains integration tests covering: missing key → skip, present key → include, `enable: false` → skip, `enable: true` → include, non-boolean → coerced + included.
- Per-plugin `enable` tests removed from `validate.test.ts` / `index.test.ts` (no longer the plugin's contract).
- `live-debugger`'s strict-rejection tests removed (no longer the contract).


watson:
* **#324** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/324?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/324?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#296** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/296?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#325](https://github.com/DataDog/build-plugins/pull/325) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/325?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

packages/plugins/apps/src/validate.test.ts
```diff
@@ -3,6 +3,22 @@
...
+import type { Logger } from '@dd/core/types';
+
+const mockLogger: Logger = {
+    getLogger: jest.fn(() => mockLogger),
+    time: jest.fn() as unknown as Logger['time'],
+    error: jest.fn(),
+    warn: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+};
```
yoannmoinet: This already exists.
https://github.com/DataDog/build-plugins/blob/00679a83a121f6d387665b4ba32d2c325507ca4d/packages/tests/src/_jest/helpers/mocks.ts#L118-L136

Same thing in the other test files.

packages/plugins/rum/README.md
yoannmoinet, line 1: This should live in its own PR.

yoannmoinet (commented):
Looks mostly good.

I wonder if we could handle one level higher than just `resolveEnable` and have something like a global `validateOptions`, that would, for now, validate `enable`.

It would be called from each plugin, but would be implemented, tested and documented just once.
Since this is a standard, this can be documented just at the root, once, and not for each individual plugin.

Or alternatively, handle this at the factory level directly, where we also have an `validateOptions` too.
https://github.com/DataDog/build-plugins/blob/00679a83a121f6d387665b4ba32d2c325507ca4d/packages/factory/src/index.ts#L176-L190

This way, plugins themselves don't even have to handle it.
The standard is applied by design.

I think I prefer the factory approach after all.

WDYT?

packages/plugins/rum/README.md
watson: Done: https://github.com/DataDog/build-plugins/pull/361

packages/plugins/apps/src/validate.test.ts
watson: Fixed 👍

jakub-g (commented):
I defer review to Yoann here :)

oliverli:
@cursor review
@codex review

oliverli (approved):
looks good for apps

packages/core/src/helpers/options.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+export const resolveEnable = <T extends { [K in C]?: unknown }, C extends string>(
+    options: T,
+    configKey: C,
+    log: Logger,
+): boolean => {
+    const pluginConfig = options[configKey];
+
+    if (pluginConfig && typeof pluginConfig === 'object' && 'enable' in pluginConfig) {
+        const value = (pluginConfig as Record<string, unknown>).enable;
```
yoannmoinet: Is this cast necessary?
I tried to remove it locally, and it wasn't failing TS.

packages/core/src/helpers/options.test.ts
```diff
@@ -0,0 +1,110 @@
...
+import { resetEnableWarnings, resolveEnable } from './options';
+
+const mockLogger: Logger = {
+    getLogger: jest.fn(),
+    time: jest.fn() as unknown as Logger['time'],
+    error: jest.fn(),
+    warn: jest.fn(),
+    info: jest.fn(),
+    debug: jest.fn(),
+};
```
yoannmoinet: A mock already exist for this.

packages/factory/src/index.test.ts
```diff
@@ -2,6 +2,17 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2019-Present Datadog, Inc.
 
+import type { PluginOptions, Options } from '@dd/core/types';
+
+const invokeFactory = async (opts: Options): Promise<PluginOptions[]> => {
+    const { buildPluginFactory } = await import('@dd/factory');
```
yoannmoinet: Any reason to have this async imported?

yoannmoinet (approved):
Looks good!
Only a few nits and questions.

packages/factory/src/index.test.ts
watson: No you're right. Fixed 👍

packages/core/src/helpers/options.test.ts
watson: Ah, found it. Fixed 👍

packages/core/src/helpers/options.ts
watson: Hmm no you're right. Not sure why it was complaining about that before. Fixed 👍

yoannmoinet (approved):
LGTM, thanks for tackling this!

PR #325, opened 2026-04-19
Validate optional live-debugger version

watson:
## Motivation

Browser Live Debugger needs a stable build identifier when we want to resolve browser probe results back to uploaded sourcemap metadata, but that value should stay optional so instrumentation can still work when no version is configured.

## Changes

This PR updates the Live Debugger build-plugin config to understand an optional `liveDebugger.version` field.

Before these changes:
- the plugin had no explicit `version` option
- there was no validation tying Live Debugger config to sourcemap upload config
- the docs did not describe how browser build identity should be configured

With these changes:
- add `version?: string` to `LiveDebuggerOptions`
- carry `version` through the validated options shape
- validate that `version` is a string when provided
- validate that `liveDebugger.version` matches `errorTracking.sourcemaps.releaseVersion` when both are configured
- keep `version` optional so Live Debugger still works without source-aware browser build lookup
- document the intended `version` semantics in the plugin README
- update unit tests and the live-debugger e2e fixture to cover the new config behavior

## Testing

- `yarn test:unit packages/plugins/live-debugger/src/validate.test.ts packages/plugins/live-debugger/src/index.test.ts`
- `ReadLints` reported no linter errors on the changed files

PR #326, opened 2026-04-21
Source build version from shared metadata.version

watson:
### What and why?

This PR makes `metadata.version` the canonical place to declare the deployed build identifier across the Datadog build plugins. Three plugins now read from it:

- **Live Debugger** injects the value into the bundle so the Browser Debugger SDK uses it as the default `version` during `init()` — keeping browser build lookup and source-code resolution working without extra SDK wiring.
- **Error Tracking** (`errorTracking.sourcemaps.releaseVersion`) falls back to it when no plugin-specific value is set.
- **RUM** (`rum.sourceCodeContext.version`) falls back to it when no plugin-specific value is set.

Users who declare the build version once at the top level no longer need to repeat it under each plugin namespace. The change is additive: existing per-plugin configurations keep working unchanged.

This builds on [#325](https://github.com/DataDog/build-plugins/pull/325), which introduced build-time validation of the version but stopped short of using it at runtime.

### How?

- `BuildMetadata` (in `@dd/core/types`) gains an optional `version?: string`, alongside the existing `name`. The root README documents it as a top-level option, and the auto-generated configuration block template in `@dd/tools` is updated to include it.
- Type-validation of `metadata.*` is centralized in the factory's `validateOptions`. Today it strictly type-checks `metadata.version`. A TODO flags `metadata.name` to be tightened in the next major — its documented default is currently `null`, so adding the check now would be a breaking change.
- The Live Debugger validator keeps a plugin-specific cross-check: `metadata.version` must match `errorTracking.sourcemaps.releaseVersion` when both are explicitly configured.
- The Live Debugger plugin builds its injected banner through a dedicated `runtime-bootstrap` helper. The bootstrap installs the minimal `$dd_probes` no-op stub, and when `metadata.version` is set it also exposes build metadata on a runtime-visible global so the Browser Debugger SDK can default its `version` from there during `init()`.
- `SourcemapsOptions.releaseVersion` is loosened from required to optional. The existing "is required" check still fires when neither `sourcemaps.releaseVersion` nor `metadata.version` resolves, with an updated message that mentions both options.

### DIFF
--- a/packages/plugins/error-tracking/src/validate.ts
+++ b/packages/plugins/error-tracking/src/validate.ts
@@ -72,36 +72,49 @@ export const validateSourcemapsOptions = (
     };
 
     if (validatedOptions.sourcemaps) {
+        const sourcemapsCfg = validatedOptions.sourcemaps;
+
+        // Resolve `releaseVersion`: prefer the plugin-specific option, then
+        // fall back to the shared top-level `metadata.version`. Letting users
+        // configure one canonical build version at the top level keeps every
+        // consumer (live-debugger, sourcemaps, …) reading from the same place.
+        const releaseVersion = sourcemapsCfg.releaseVersion || config.metadata?.version;
+
         // Validate the configuration.
-        if (!validatedOptions.sourcemaps.releaseVersion) {
-            toReturn.errors.push(`${red('sourcemaps.releaseVersion')} is required.`);
+        if (!releaseVersion) {
+            toReturn.errors.push(
+                `${red('sourcemaps.releaseVersion')} is required (set it directly or via ${red('metadata.version')}).`,
+            );
         }
-        if (!validatedOptions.sourcemaps.service) {
+        if (!sourcemapsCfg.service) {
             toReturn.errors.push(`${red('sourcemaps.service')} is required.`);
         }
-        if (!validatedOptions.sourcemaps.minifiedPathPrefix) {
+        if (!sourcemapsCfg.minifiedPathPrefix) {
             toReturn.errors.push(`${red('sourcemaps.minifiedPathPrefix')} is required.`);
         }
 
         // Validate the minifiedPathPrefix.
-        if (validatedOptions.sourcemaps.minifiedPathPrefix) {
-            if (!validateMinifiedPathPrefix(validatedOptions.sourcemaps.minifiedPathPrefix)) {
-                toReturn.errors.push(
-                    `${red('sourcemaps.minifiedPathPrefix')} must be a valid URL or start with '/'.`,
-                );
-            }
+        if (
+            sourcemapsCfg.minifiedPathPrefix &&
+            !validateMinifiedPathPrefix(sourcemapsCfg.minifiedPathPrefix)
+        ) {
+            toReturn.errors.push(
+                `${red('sourcemaps.minifiedPathPrefix')} must be a valid URL or start with '/'.`,
+            );
         }
 
-        // Add the defaults.
-        const sourcemapsWithDefaults: SourcemapsOptionsWithDefaults = {
-            bailOnError: false,
-            dryRun: false,
-            maxConcurrency: 20,
-            ...validatedOptions.sourcemaps,
-        };
-
-        // Save the config.
-        toReturn.config = sourcemapsWithDefaults;
+        // Build the resolved config only when `releaseVersion` actually
+        // resolves; otherwise an error has been recorded and the caller will
+        // throw before the config is read.
+        if (releaseVersion) {
+            toReturn.config = {
+                bailOnError: false,
+                dryRun: false,
+                maxConcurrency: 20,
+                ...sourcemapsCfg,
+                releaseVersion,
+            };
+        }
     }
 
     return toReturn;
--- a/packages/plugins/error-tracking/src/validate.ts
+++ b/packages/plugins/error-tracking/src/validate.ts
@@ -86,6 +86,15 @@ export const validateSourcemapsOptions = (
                 `${red('sourcemaps.releaseVersion')} is required (set it directly or via ${red('metadata.version')}).`,
             );
         }
+        if (
+            sourcemapsCfg.releaseVersion &&
+            config.metadata?.version &&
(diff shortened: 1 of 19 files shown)


watson:
* **#337** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/337?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#335** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/335?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#326** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/326?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/326?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tylfin (approved):

packages/plugins/live-debugger/README.md
```diff
@@ -126,12 +127,22 @@ const double = (x) => {
...
+
+```ts
+datadogBuildPlugins({
+    metadata: { version: '1.0.0' },
+    liveDebugger: {},
+});
+```
 
-Optional. When set, use an immutable deployed browser build identifier. This value should match:
+If both `metadata.version` and an explicit `errorTracking.sourcemaps.releaseVersion` are configured and disagree, this plugin surfaces the mismatch as a build error.
```
yoannmoinet: Not sure this is super relevant for the Live Debugger plugin's documentation.

packages/plugins/live-debugger/src/validate.ts
```diff
@@ -21,17 +22,10 @@ export const validateOptions = (config: Options, log: Logger): LiveDebuggerOptio
...
-    if (
-        pluginConfig.version &&
-        sourcemapReleaseVersion &&
-        pluginConfig.version !== sourcemapReleaseVersion
-    ) {
+    // Validate version
+    if (metadataVersion && sourcemapReleaseVersion && metadataVersion !== sourcemapReleaseVersion) {
         errors.push(
-            `${red('version')} must match ${red('errorTracking.sourcemaps.releaseVersion')} when both Live Debugger and sourcemap upload are configured`,
+            `${red('metadata.version')} must match ${red('errorTracking.sourcemaps.releaseVersion')} when both Live Debugger and sourcemap upload are configured`,
```
yoannmoinet: Does it make sense to have this validation in the live debugger plugin?

I would probably see this more in the error tracking plugin instead.

packages/plugins/live-debugger/src/validate.test.ts
```diff
@@ -17,8 +17,8 @@ const mockLogger: Logger = {
     debug: jest.fn(),
 };
 
-const makeConfig = (liveDebugger?: unknown, errorTracking?: unknown): Options =>
-    ({ liveDebugger, errorTracking }) as unknown as Options;
+const makeConfig = (liveDebugger?: unknown, errorTracking?: unknown, metadata?: unknown): Options =>
+    ({ liveDebugger, errorTracking, metadata }) as unknown as Options;
```
yoannmoinet: Nit: not a fan of all these `unknown` casting.
All the types are accessible and we have a bunch of mocks to alleviate the burden of mocking these out.

packages/plugins/error-tracking/src/validate.ts
```diff
@@ -72,36 +72,49 @@ export const validateSourcemapsOptions = (
...
-            maxConcurrency: 20,
-            ...validatedOptions.sourcemaps,
-        };
-
-        // Save the config.
-        toReturn.config = sourcemapsWithDefaults;
+        // Build the resolved config only when `releaseVersion` actually
+        // resolves; otherwise an error has been recorded and the caller will
+        // throw before the config is read.
+        if (releaseVersion) {
```
yoannmoinet, line 118: Not sure this `if () {}` wrapping is really necessary.
It brings more confusion when reading the code than anything.

Even the comment tends to explain that it's actually unnecessary since it would throw anyways.

packages/plugins/live-debugger/src/types.ts
```diff
@@ -15,7 +15,6 @@ export type FunctionKind = (typeof VALID_FUNCTION_KINDS)[number];
 
 export type LiveDebuggerOptions = {
     enable?: boolean;
-    version?: string;
```
yoannmoinet, line 18: I believe it's acceptable to remove it from the public API since the plugin didn't actually release publicly yet, but just to confirm it's not been missed.

yoannmoinet (commented):
I'm not sure I understand why the `errorTracking` concern is handled by the `liveDebugger` plugin.

I don't think these two plugins should know anything about each others.

packages/plugins/error-tracking/src/validate.ts
watson: I think it's best to keep the code as-is because `releaseVersion` might be `undefined` and if we don't have this if-guard, we get a TypeScript error below because it's not allowed to be `undefined` in that context. If you have a better fix for this, let me know.

packages/plugins/live-debugger/src/types.ts
watson: Yes, this is intentional. The Live Debugger plugin has not shipped publicly yet, so I removed the plugin-specific `liveDebugger.version` option and now read the build version only from the shared top-level `metadata.version`.

packages/plugins/live-debugger/README.md
watson: Agree - removed 👍

packages/plugins/live-debugger/src/validate.test.ts
watson: Fixed 👍

packages/plugins/live-debugger/src/validate.ts
watson: No, you're probably right. Fixed 👍

yoannmoinet (approved):
LGTM

PR #331, opened 2026-04-27
[APPS] Add backend function connection support

sdkennedy2:
## Motivation

High Code App backend functions today **fail at runtime** when they call an action that requires a `connectionId` (e.g. `@datadog/action-catalog/openai/openai#chatCompletion`). The Action Platform server rejects the call with `Access denied for connection {uuid}. Add it to the allowed list in the step configuration.` because the build plugin never tells the server which connections each function is allowed to use.

This PR implements the [Backend Function Connections RFC](https://datadoghq.atlassian.net/wiki/spaces/HCA/pages/6629949443/RFC+Backend+Function+Connections), which chose a **single-file** approach: customers maintain a centralized `connections.ts` at the project root listing every connection their app uses, and the build plugin reads it once at build time to feed both the production upload manifest and the dev-preview request body.

CLI tooling and the ESLint rule from the RFC's Future Enhancements section are intentionally out of scope.

### Customer-facing surface

Customers add a `connections.ts` (or `.tsx` / `.js` / `.jsx`) at the project root. Either `CONNECTIONS` (uppercase) or `connections` (lowercase) is accepted as the export name:

```ts
// connections.ts
export const CONNECTIONS = {
    OPEN_AI: '9e51df63-7cba-4067-a1fc-a156195074e8',
    SLACK:   '321d6080-5c01-4793-be58-475f0f20e801',
} as const;
```

Backend functions reference connection IDs as before (literal UUIDs or via `CONNECTIONS.OPEN_AI` once they import it). On `npm run build`, the plugin emits a `manifest.json` at the **root of the upload ZIP** with the per-function allowed connection IDs nested under `backend.functions`, keyed by encoded query name. Every backend function key gets the same `allowedConnectionIds` array (the union from `connections.ts`). The server-side actions runtime supports per-function distinct lists, but the RFC explicitly chose the flat union.

```json
{
  "backend": {
    "functions": {
      "631339bbf278cb23b10f1118572d0e3d21166de9e4363c6941d055c9f1b19ea1.helloBackend": {
        "allowedConnectionIds": [
          "321d6080-5c01-4793-be58-475f0f20e801",
          "9e51df63-7cba-4067-a1fc-a156195074e8"
        ]
      }
    }
  }
}
```

For dev preview, the same list is forwarded inside `inputs.allowedConnectionIds` of the `/api/v2/app-builder/queries/preview-async` request body, matching the action shape the server expects.

## QA Instructions

End-to-end against a scaffolded high-code-app (e.g. `~/dd/web-ui/test-action-catalog-app`):

1. **No connections file, build path** — remove any `connections.ts`, run `npm run build` with `logLevel: 'debug'`. Expect `manifest.json` at the root of the upload ZIP with `backend.functions[<key>].allowedConnectionIds: []` for every function.
2. **With connections file, build path** — add `connections.ts` exporting two UUIDs, `npm run build`, inspect the assembled ZIP — root `manifest.json` should contain every backend-function key under `backend.functions`, each with both UUIDs in `allowedConnectionIds`.
3. **Dev preview** — with `DD_API_KEY` / `DD_APP_KEY` set and a `connections.ts` containing the OpenAI UUID, add a backend function calling `chatCompletionAction({ ..., connectionId: '...' })`. `npm run dev`, click the action — request should succeed (no `Access denied for connection ...`). Edit `connections.ts` to remove the UUID, re-trigger — should fail with the expected denial. Verifies `handleHotUpdate`-driven re-extraction.
4. **Lowercase variant** — rename the export to `export const connections = { ... }`, repeat steps 2–3. Behavior should be identical.
5. **Strict failure modes** — `OPEN_AI: process.env.OPEN_AI_ID` → `npm run build` fails with a framed error pointing at the offending node. `default export` form → fails with `connections file must define "export const CONNECTIONS" (or "connections") = { ... }`.
6. **Unit tests** — `yarn test:unit packages/plugins/apps` (171 tests pass, including 26 extractor tests covering both export-name variants, buildStart/handleHotUpdate test groups, and a dev-server `allowedConnectionIds` assertion).
7. **E2E tests** — `yarn test:e2e packages/tests/src/e2e/appsPlugin` (60 tests pass, including the new root `manifest.json` shape assertion).

## Blast Radius

- Only affects customers building high-code apps (`@datadog/vite-plugin` with `apps.enable`). Other bundlers and other plugin features are untouched.
- **Strict failure mode:** unresolvable values inside `connections.ts` now fail the build where they previously didn't. Customers without a `connections.ts` see no change — the manifest is emitted with empty `allowedConnectionIds` arrays, matching pre-existing behavior (no connection-using actions allowed).
- The dev preview now sends `allowedConnectionIds` inside the `inputs` field of the preview-async request body. This is additive — empty list means same behavior as today.
- Shipping path: this is the first PR implementing the RFC. CLI helper and ESLint rule are deferred to follow-ups per RFC Future Enhancements.

## Documentation

-   [RFC: Backend Function Connections](https://datadoghq.atlassian.net/wiki/spaces/HCA/pages/6629949443/RFC+Backend+Function+Connections)
-   [App Builder Backend Functions Calling External Services](https://datadoghq.atlassian.net/wiki/spaces/~61b10ff3c75da800726413a5/pages/6609798666)


sdkennedy2:
* **#332** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/332?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#331** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/331?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/331?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review

packages/plugins/apps/src/backend/extract-connections.test.ts
```diff
@@ -0,0 +1,437 @@
...
+
+    test('Should throw when both "connections" and "CONNECTIONS" are exported', () => {
+        const ast = program([
+            exportConnections([stringProperty('A', 'uuid-1')], 'connections'),
+            exportConnections([stringProperty('B', 'uuid-2')], 'CONNECTIONS'),
+        ]);
+        expect(() => extractConnectionIds(ast, filePath, '')).toThrow(
+            'multiple top-level "export const CONNECTIONS" (or "connections") declarations are not allowed',
+        );
+    });
```
sarenji: Out of curIosity, why should we support both cases? Do we need to, or can we pick one?

packages/plugins/apps/src/backend/extract-connections.test.ts
```diff
@@ -0,0 +1,437 @@
...
+        expect(() => extractConnectionIds(ast, filePath, '')).toThrow('computed keys');
+    });
+
+    test('Should throw on spread elements', () => {
+        const ast = program([
+            exportConnections([
+                {
+                    type: 'SpreadElement',
+                    argument: { type: 'Identifier', name: 'other' },
+                } as SpreadElement,
```
sarenji: Does `satisfies` work here in place of `as`?

packages/plugins/apps/src/backend/extract-connections.test.ts
```diff
@@ -0,0 +1,437 @@
...
+        },
+    );
+
+    // Rollup's this.parse() (SWC) emits character offsets but no line:col,
+    // so we derive line:col from `node.start` against the source text.
+    // These tests use the real parser so a regression in offset handling
+    // surfaces immediately.
+    describe('framed source location from parseAst', () => {
+        test('Should include line:col when value is not a string literal', async () => {
+            const { parseAst } = await import('rollup/parseAst');
```
sarenji: Why is this dynamically imported? Also why rollup's AST parser specifically? I feel like something should expose the parser somewhere that we can use instead of doing this.

packages/plugins/apps/src/vite/dev-server.test.ts
```diff
@@ -94,13 +94,14 @@ describe('Dev Server Middleware', () => {
...
-            mockLog,
-        );
+        const middleware = createDevServerMiddleware({
+            viteBuild: mockViteBuild,
+            getBackendFunctions: () => mockFunctions,
+            getConnectionIds: () => [],
+            auth: mockAuth,
+            projectRoot: '/project',
+            log: mockLog,
+        });
```
sarenji, line 104: Nice, this is a much better interface

packages/plugins/apps/src/vite/dev-server.test.ts
```diff
@@ -375,6 +378,54 @@ describe('Dev Server Middleware', () => {
...
+
+            const inputs = (
+                capturedBody as {
+                    data: {
+                        attributes: { query: { properties: { spec: { inputs: unknown } } } };
+                    };
+                }
+            ).data.attributes.query.properties.spec.inputs as {
+                allowedConnectionIds: string[];
+            };
```
sarenji: This is an awkward construction especially when `capturedBody` is set to `unknown` and `inputs` is also set to `unknown`

packages/plugins/apps/src/vite/index.ts
```diff
@@ -15,48 +17,133 @@ export interface VitePluginOptions {
...
 
-/**
- * Returns the Vite-specific plugin hooks for the apps plugin.
- *
- * Production (closeBundle): builds backend functions (if any) then uploads
- * all assets sequentially.
- *
- * Dev (configureServer): registers middleware for local backend function
- * testing when auth credentials are available.
- */
```
sarenji, line 31: Please restore this comment

packages/plugins/apps/src/index.test.ts
sarenji, line 1: There's a lot of mocking and extracting in the new helper functions at the top... is there a way to simplify without faking everything? The more mocks/fakes there are, the less confidence I have in the code

packages/plugins/apps/src/vite/index.ts
```diff
@@ -15,48 +17,133 @@ export interface VitePluginOptions {
...
+        // apps plugin, so its buildStart can't help).
+        async buildStart() {
+            connectionRegistry.setParse((code) => this.parse(code));
+            try {
+                const { filePath } = await connectionRegistry.loadAndSetConnectionIds(
+                    async (id) => {
+                        if (devServer) {
+                            const result = await devServer.transformRequest(id);
+                            return result?.code ?? null;
+                        }
```
sarenji, line 71: Can you explain why `this.load` is not sufficient and you need to capture `devServer` from a different hook? (What does `this.load` do vs `devServer.transformRequest`?)

And if it's really necessary, can the reason be captured in a code comment?

packages/tests/src/e2e/appsPlugin/appsPlugin.spec.ts
```diff
@@ -160,14 +160,31 @@ describe('Apps Plugin', () => {
...
+        const manifest = JSON.parse(await manifestFile!.async('string')) as {
+            backend: { functions: Record<string, { allowedConnectionIds: string[] }> };
+        };
+        const backendFiles = filePaths.filter((f) => f.startsWith('backend/') && !zip.files[f].dir);
+        const expectedKeys = backendFiles.map((f) =>
+            f.replace(/^backend\//, '').replace(/\.js$/, ''),
+        );
+        expect(Object.keys(manifest.backend.functions).sort()).toEqual(expectedKeys.sort());
+        for (const entry of Object.values(manifest.backend.functions)) {
+            expect(Array.isArray(entry.allowedConnectionIds)).toBe(true);
```
sarenji: Are we able to test the values of the array as well, or are we satisfied with an `Array.isArray` check?

sarenji (commented):
Some thoughts

packages/plugins/apps/src/vite/index.ts
sdkennedy2: I actually intentionally removed it because I thought it was weird describing all of the inner functions at the top (claude wanted to add even more to it as the complexity of the plugin grew), but it could be useful to differentiate the production versus dev. I'm not particularly opinionated. I can add it back.

packages/plugins/apps/src/vite/index.ts
sdkennedy2: Essentially this.load doesn't work in devServer mode. I can add a comment.

packages/plugins/apps/src/index.test.ts
sdkennedy2: I wonder if we should just exclude index.ts from unit tests and rely on the e2e test. This file mostly just glues everything together and it might just be better to verify everything works end to end.

We could in a later PR just refactor index.ts and pull out distinct pieces and create unit tests for them directly for example:
- Upload function
- Registry functions
- etc

packages/plugins/apps/src/backend/extract-connections.test.ts
sdkennedy2: Thats fair pushback. I just thought it was an easy foot gun to make. I plan on adding a couple eslint rules perhaps we standardize on the uppercase version and we can avoid the footgun with linting.

packages/plugins/apps/src/vite/dev-server.test.ts
sdkennedy2: That is wild. Claude wrote that. I glossed over the tests, I usually don't invest too much resources in unit test coverage during the design partner phase since nothing is really set in stone and things are likely to change.

This line is pretty egregious though haha. Happy to fix it.

packages/plugins/apps/src/backend/extract-connections.ts
```diff
@@ -0,0 +1,173 @@
...
+ */
+export async function findConnectionsFile(buildRoot: string): Promise<string | undefined> {
+    for (const ext of CONNECTIONS_EXTENSIONS) {
+        const candidate = path.join(buildRoot, `${CONNECTIONS_FILE_BASENAME}${ext}`);
+        try {
+            await fsp.access(candidate);
+            return candidate;
+        } catch {
+            // not found at this extension — try the next.
+        }
```
yoannmoinet, line 27: This will swallow any errors, not just the "not found" errors.
Maybe a specific gate of `ENOENT` errors only would be necessary.
Unless we're ok with this.

yoannmoinet (approved):
Overall looks good to me.
Just a nit on some error handling that can be misleading.

sdkennedy2:
@codex review
@cursor review

PR #353, opened 2026-05-10
[APPS][Connections Part 6] Build backend module graph during backend builds

sdkennedy2:
## Motivation

Prepare backend connection ID analysis to use the same module graph that backend bundling already sees, without moving connection ID extraction into this PR.

## Changes

This PR adds module graph collection to the backend build path used by both production builds and the dev server. During each backend build, the Vite collector observes Rollup `moduleParsed` events and records app-local, parseable source modules with their transformed ASTs, canonical module IDs, static dependency IDs, and unsupported local dependency facts.

For example, this backend module:

```ts
// /app/src/backend/actions.backend.ts
import { getMessage } from "./helpers";

export async function run() {
    return getMessage();
}
```

Produces a collected module record shaped like this:

```ts
{
    id: "/app/src/backend/actions.backend.ts",
    ast: Program,
    staticDependencies: ["/app/src/backend/helpers.ts"],
    unsupportedDependencies: [],
}
```

The Vite-specific adapter owns Rollup module ID normalization and virtual module filtering, while the backend AST module graph code works with canonical app-local module IDs. That keeps bundler details out of the reusable graph record layer.

This PR intentionally stops at building the graph. Traversing the graph to collect connection IDs from reachable modules is handled in the next PR in the stack. Existing same-file backend connection ID behavior remains unchanged.

## QA Instructions

Updated the focused unit coverage for backend module graph record creation and Vite collector behavior. Also ran the apps plugin typecheck and lint coverage for the changed module graph and backend build integration paths.

Validated the change against a local test app to confirm backend functions still build and upload while existing same-file `allowedConnectionIds` behavior remains unchanged.

## Blast Radius

This affects the apps plugin backend build path by collecting module graph metadata during backend builds. The collected graph is not used to change emitted `allowedConnectionIds` behavior in this PR, so runtime behavior should remain unchanged.

## Documentation

- [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)

PR #351, opened 2026-05-08
[APPS][Connections Part 3] Resolve same-module connection ID constants

sdkennedy2:
## Motivation

This builds on the inline connection ID extractor from #349. Inline string literals are safe, but backend code commonly assigns a connection ID to a same-file constant before passing it to action-catalog.

This PR adds the smallest useful same-module value resolver first so the follow-up object-map PR can be reviewed separately.

## Changes

Adds same-module static value resolution for direct constant forms:

```ts
const CONNECTION_ID = '77c14b8b-27e1-4901-985d-8817908b9706';
request({ connectionId: CONNECTION_ID, inputs: {} });
```

The resolver tracks declarations by `eslint-scope` variable identity, preserving shadowing behavior from #349. It supports top-level same-file const strings, exported top-level consts, const-to-const chains, and static template literals without interpolation.

It intentionally rejects member expressions such as `CONNECTIONS.HTTP`, imported values, mutable bindings, function-local bindings, dynamic templates, calls, binary expressions, environment reads, and const cycles. Same-module object-map support is split into the next PR.

## QA Instructions

Ran:

```sh
yarn workspace @dd/tests test:unit packages/plugins/apps/src/backend/ast-parsing/extract-connection-ids.test.ts --runInBand
```

Parent branch result: 46 tests passed.

## Blast Radius

This affects Apps backend function discovery/upload/dev-server manifest generation for action-catalog connection allowlists. Runtime backend execution is unchanged. Unsupported known action-catalog `connectionId` forms fail closed instead of producing an incomplete allowlist.

## Documentation

- [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)

PR #354, opened 2026-05-11
[APPS][Connections Part 4] Resolve same-module connection ID object values

sdkennedy2:
## Motivation

The parent PR adds direct same-file connection ID constants. App code also commonly groups same-file connection IDs in local object maps, for example `CONNECTIONS.HTTP` or `CONNECTIONS.HTTP.PROD`.

This PR adds that object-map support separately from the basic constant resolver so reviewers can evaluate the property-access rules on their own.

## Changes

Adds same-module object member resolution for supported `connectionId` forms:

```ts
const CONNECTIONS = { HTTP: '77c14b8b-27e1-4901-985d-8817908b9706' };
request({ connectionId: CONNECTIONS.HTTP, inputs: {} });
```

The algorithm works in two steps.

First, it treats `CONNECTIONS.HTTP` as a static object lookup. It resolves `CONNECTIONS` to the same-file const object, finds the `HTTP` property, and returns the expression stored at that property. In the example above, that expression is the string literal `'77c14b8b-27e1-4901-985d-8817908b9706'`.

That property value does not need to be a string literal directly. It can also be another same-file const identifier:

```ts
const HTTP_CONNECTION_ID = '77c14b8b-27e1-4901-985d-8817908b9706';
const CONNECTIONS = { HTTP: HTTP_CONNECTION_ID };
request({ connectionId: CONNECTIONS.HTTP, inputs: {} });
```

In that case, the object lookup returns the identifier expression `HTTP_CONNECTION_ID`.

Second, the returned expression goes back through the existing connection ID value resolver. If the expression is a string literal, it becomes the final connection ID immediately. If the expression is an identifier, the resolver follows that identifier to its same-file const initializer and then resolves that initializer to the final string.

This same two-step shape also supports nested static reads such as `CONNECTIONS.HTTP.PROD`: each object member lookup returns the next expression, and the general resolver keeps resolving until it reaches a final supported string value.

The resolver intentionally fails closed for dynamic forms that could hide the runtime value, including computed member reads, imported object roots, mutable object roots, object spreads, computed object properties, duplicate keys, accessors, missing properties, and non-object intermediate values. Imported connection ID tracing remains out of scope for a later PR.

## QA Instructions

Ran:

```sh
yarn workspace @dd/tests test:unit packages/plugins/apps/src/backend/ast-parsing/extract-connection-ids.test.ts --runInBand
```

Child branch result: 58 tests passed.

## Blast Radius

This affects Apps backend function discovery/upload/dev-server manifest generation for action-catalog connection allowlists. Runtime backend execution is unchanged. The resolver remains conservative and throws on object shapes that could hide the actual runtime connection ID.

## Documentation

- [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)

### DIFF
--- a/packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
+++ b/packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
@@ -7,6 +7,7 @@ import type {
     Expression,
     Identifier,
     Literal,
+    MemberExpression,
     ObjectExpression,
     Program,
     Property,
@@ -101,8 +102,8 @@ type StaticBinding =
  *   `connectionId: HTTP` fails closed instead of trusting a value that can
  *   change.
  * - `unsupported-pattern`: remember that the declaration came from a binding
- *   pattern such as `const { HTTP } = CONNECTIONS`, which this PR intentionally
- *   does not resolve.
+ *   pattern such as `const { HTTP } = CONNECTIONS`, which this resolver
+ *   intentionally does not support.
  */
 export interface SameModuleConnectionIdBindings {
     byVariable: Map<eslintScope.Variable, StaticBinding>;
@@ -130,7 +131,7 @@ interface ConnectionIdResolutionContext {
  * `const HTTP_CONNECTION_ID = 'abc'`. Top-level `let` and `var` declarations are
  * recorded as mutable so they fail closed if used. Destructured declarations are
  * recorded as unsupported because `const { HTTP } = CONNECTIONS` adds more
- * aliasing behavior than this PR intends to support.
+ * aliasing behavior than the resolver supports.
  */
 export function collectSameModuleConnectionIdBindings(
     ast: Program,
@@ -141,6 +142,7 @@ export function collectSameModuleConnectionIdBindings(
     for (const node of ast.body) {
         // Top-level declarations such as:
         //   const HTTP_CONNECTION_ID = 'abc';
+        //   const CONNECTIONS = { HTTP: 'abc' };
         if (node.type === 'VariableDeclaration') {
             collectVariableDeclarationBindings(node, scopeAnalysis, byVariable);
             continue;
@@ -214,7 +216,8 @@ function collectVariableDeclarationBindings(
     for (const declarator of declaration.declarations) {
         const variables = scopeAnalysis.scopeManager.getDeclaredVariables(declarator);
 
-        // Destructuring creates variables, but we do not resolve it in this PR:
+        // Destructuring creates variables, but this resolver does not follow
+        // destructured aliases:
         //   const { HTTP } = CONNECTIONS;
         if (declarator.id.type !== 'Identifier') {
             for (const variable of variables) {
@@ -233,6 +236,7 @@ function collectVariableDeclarationBindings(
 
         // Immutable top-level values can be followed later:
         //   const HTTP_CONNECTION_ID = 'abc';
+        //   const CONNECTIONS = { HTTP: HTTP_CONNECTION_ID };
         if (declaration.kind === 'const') {
             byVariable.set(variable, { kind: 'const', init: declarator.init ?? null });
         } else {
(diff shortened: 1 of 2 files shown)


sdkennedy2:
* **#362** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/362?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#363](https://github.com/DataDog/build-plugins/pull/363) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/363?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#365](https://github.com/DataDog/build-plugins/pull/365) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/365?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#364** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/364?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#353** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/353?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#357** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/357?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#354** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/354?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/354?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#351** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/351?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
```diff
@@ -365,6 +411,228 @@ function resolveIdentifierValue(
...
+    }
+
+    context.seen.add(variable);
+    try {
+        const objectExpression = resolveObjectExpressionValue(binding.init, context);
+        // `CONNECTIONS.HTTP` only works when `CONNECTIONS` is visibly an object
+        // literal in this file, directly or through const aliases:
+        //   const CONNECTIONS = { HTTP: 'abc' };
+        //   const ACTIVE_CONNECTIONS = CONNECTIONS;
+        return objectExpression;
```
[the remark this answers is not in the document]
sdkennedy2: I'm fine with this being allowed. While the object could be mutated I think the other options are either too restrictive or a lot of work to detect mutations.

packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
```diff
@@ -399,6 +406,228 @@ function resolveIdentifierValue(
...
+ * Example:
+ *
+ * ```ts
+ * const CONNECTIONS = { HTTP: 'abc' };
+ * request({ connectionId: CONNECTIONS.HTTP });
+ * ```
+ *
+ * This supports static property chains such as `IDENTIFIER.STATIC.PROPERTY`.
+ * Computed reads like `CONNECTIONS[key]` and imported objects are rejected
+ * because they need more analysis than this PR owns.
```
sdkennedy2: We should not check in comments that reference the current PR. The code lives outside the context of a single PR.

packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
```diff
@@ -400,6 +408,240 @@ function resolveIdentifierValue(
...
+    }
+    if (binding.kind === 'mutable') {
+        // A mutable map can change before the action call runs:
+        //   let CONNECTIONS = { HTTP: 'abc' };
+        //   CONNECTIONS = loadConnections();
+        throw unsupportedConnectionId(
+            context.filePath,
+            `mutable ${binding.declarationKind} connectionId object binding ${node.name}`,
+        );
+    }
```
sarenji, line 532: Just to be sure, is this mutable feature something we could realistically see a user wanting in the future?

sarenji (approved):
🔥

packages/plugins/apps/src/backend/ast-parsing/connection-id-values.ts
sdkennedy2: We can always iterate on customer feedback and prioritize trying to support this if our customers ask for it.

PR #360, opened 2026-05-12
fix: replace Node.js pipe() with Web Streams pipeThrough() to fix NotReadableError flake

yoannmoinet:
## Summary

Fixes a long-standing flaky test failure (`NotReadableError: The blob could not be read`) in the unit test suite.

### Root Cause

The error is a race condition in `createRequestData()` in `packages/core/src/helpers/request.ts`.

**The chain of events:**
1. `getData()` builds FormData with `fs.openAsBlob()`-backed blobs (lazy file references)
2. `createRequestData()` serialized the form through a `Request`, then used Node.js `Readable.fromWeb().pipe(createGzip())` to compress it
3. `pipe()` immediately calls `.resume()` on the source stream, scheduling actual blob reads via `process.nextTick()`
4. In tests, nock intercepts the `fetch()` call and replies **without consuming the request body** — the async chain completes almost instantly
5. The test finishes → cleanup runs → `rm(rootDir)` deletes the output directory
6. `process.nextTick()` fires, tries to read the now-deleted file-backed blob → `NotReadableError`
7. The unhandled stream error gets picked up by dd-trace, which misattributes it to the `test.each` registration context, making it appear as a flaky failure on a random test

### Fix

Replace `Readable.fromWeb().pipe(createGzip())` with the Web Streams `pipeThrough(new CompressionStream('gzip'))`. Web Streams are **lazy**: the pipeline only starts reading when the output stream is actually consumed (i.e., when `fetch` reads the body). Since nock intercepts without reading the body, no blob I/O ever starts, eliminating the race.

This also simplifies the code by removing the Node.js `stream` and `zlib` imports in favor of the Web Streams `CompressionStream` (available since Node.js 18).

## Changes

- `packages/core/src/helpers/request.ts`: Replace `Gzip | Readable` with `ReadableStream`, use `CompressionStream` instead of `createGzip`, remove Node.js stream/zlib imports
- `packages/plugins/error-tracking/src/sourcemaps/sender.test.ts`: Update `readFully()` helper to use Web Streams API to match the new `ReadableStream` type

## Test plan

- [ ] `yarn test:unit packages/plugins/error-tracking` — all 9 sender tests pass
- [ ] `yarn test:unit packages/core` — all 11 request helper tests pass
- [ ] Run the integration test suite a few times to confirm the flake is gone


yoannmoinet:
@codex review
@cursor review

sdkennedy2 (approved):

PR #144, opened 2025-02-27
[rum] Add RUM's plugin documentation

yoannmoinet:
### What and why?

Make the feature visible.

### How?

Adding back its `README` and remove it from the list of README exceptions.

### TODO

- [x] Find a way to bundle in types in the plugin.

PR #370, opened 2026-05-15
Replace rollup-plugin-dts with dts-bundle-generator

yoannmoinet:
### What and why?

Replace `rollup-plugin-dts` with `dts-bundle-generator` for bundling the published packages' TypeScript declarations. The previous setup hit DOM-lib vs `@types/node` conflicts when bundling types that pulled in browser APIs (e.g. the RUM Browser SDK types) and was flagged as slow in the rollup config.

This unblocks #144 (exposing the RUM plugin in the root README and dropping its `hideFromRootReadme` flag). #144 wants consumers of the published bundler plugins to be able to configure RUM without installing `@datadog/browser-rum` themselves — which requires `RumInitConfiguration` and `RumPublicApi` to be bundled into the shipped `.d.ts`. Under `rollup-plugin-dts` the bundle failed on DOM-vs-Node lib conflicts the moment those SDK types entered the graph, so #144 carried a `// TODO: Find a way to bundle in types in the plugin.` and a fallback path that re-declared the SDK types by hand. With this PR's two-pass + `inlinedLibraries` setup, the SDK types are inlined cleanly into the published `.d.ts`, and #144 can drop the workaround and use `Assign<RumInitConfiguration, …>` directly.

#### Two-pass bundling (`packages/tools/src/dtsBundlePlugin.mjs`)
- **Pass 1** — TypeScript API emits `.d.ts` for every workspace file reachable from the entry, using the root tsconfig (no DOM lib, so no `@types/node` conflict).
- **Pass 2** — `dts-bundle-generator` bundles against the emitted `.d.ts` files with DOM lib enabled. Because every reachable file is now a declaration (entry + `@dd/*` redirected via `paths`), dts-bundle-generator hits its `allFilesAreDeclarations` shortcut and skips its own compilation — so the lib conflict never resurfaces.
- Temp artifacts (emitted `.d.ts`, the bundle tsconfig) land in `<CWD>/.dts-tmp-<package>/` so TypeScript resolves external packages via the project's `node_modules` by walking up — no symlink hacks. The whole `mkdir → emit → bundle → write` flow is wrapped in `try/finally`, so a failure cleans up.
- Lives in its own module so the dts-specific deps (`dts-bundle-generator`, `typescript`) don't leak into `rollupConfig.mjs`'s import graph.

#### Workspace `@dd/*` paths
- Declared once in the root `tsconfig.json` (`compilerOptions.paths`). Both passes extend the root config and inherit them; pass 2 rewrites them to point at the emitted `.d.ts` files inside the temp dir.
- A new `yarn cli integrity` check derives the expected mapping from each workspace's `package.json` `exports` and reconciles it with the root tsconfig — idempotent when correct, auto-fixes drift, always writes alphabetically sorted.

#### Plugin `buildPlugin` config
- Each plugin's `package.json` now carries its build-time config under a single `buildPlugin` block (alongside the pre-existing `hideFromRootReadme`):
  - `buildPlugin.inlinedLibraries`: npm packages whose types should be folded into the bundled `.d.ts` rather than left as imports (currently `@datadog/browser-core` + `@datadog/browser-rum-core`, declared on `@dd/rum-plugin`).
  - `buildPlugin.toBuild`: sub-build descriptors that used to live at the top level of `package.json`.
- The dts plugin unions every plugin's `inlinedLibraries` at build time — no shared rollup-config edits when adding a future plugin that needs to inline another library.

#### Library configuration at the dts-bundle-generator call site
- `inlinedLibraries`: union of plugin-declared values.
- `importedLibraries`: every peer/runtime dep that isn't inlined — preserved as `import` statements.
- `output.exportReferencedTypes: false` — keeps internally-referenced SDK types (`Site`, `BufferedObservable`, `Encoder`, …) as `declare` rather than `export`. Without this, dts-bundle-generator's default behaviour was leaking ~280 internal types as part of our public API surface; the documented surface is now ~10 names (`VitePluginOptions`, `datadogVitePlugin`, the per-plugin `*Types` lookups, `helpers`, `version`).

#### RUM SDK types
- `packages/plugins/rum/src/browserSdkTypes.ts` re-exports `RumPublicApi` and `RumInitConfiguration` from `@datadog/browser-rum-core` directly (instead of the top-level `@datadog/browser-rum`). This matches what's inlined and avoids a `typeof datadogRum` indirection that confused the bundler.

#### Yarn patch on dts-bundle-generator
- Adds a null check in `TypesUsageEvaluator` for cases where `getImportExportReferencedSymbol` returns `undefined` (triggered by `postcss.d.mts` re-exporting from `./postcss.js` under bundler-mode resolution).
- Verified empirically that the patch is still required even without `inlinedLibraries`: the evaluator walks types from `importedLibraries` and other transitive sources too, so the bug is reachable through `vite` / `rollup` peer-dep type graphs.


michaeldfoley (approved):

amortemousque (approved):

yoannmoinet:
@codex review
@cursor review

PR #372, opened 2026-05-15
Constrain auth.site to the documented Datadog sites

oliverli:
### What and why?

Previously `AuthOptions.site` was string, which let typos flow silently into Datadog intake URLs.

It was also a major source of bugs as sometimes Datadog API endpoints would inadvertently be specified as e.g. `https://${auth.site}/api/etc/etc` without the `api` subdomain, causing 403s when the APIs are run.

### How?

- Introduces a SITES runtime tuple in packages/core/src/constants.ts as the
  single source of truth (matches the existing ALL_ENVS / SUPPORTED_BUNDLERS
  pattern).
- Derives the Sites type from SITES so type and runtime stay in lockstep.
- Validates auth.site and DATADOG_SITE / DD_SITE in validateOptions and
  throws a clear error pointing at
  https://docs.datadoghq.com/getting_started/site/ when an unsupported value
  is provided.
- Renames the test helper FAKE_SITE -> DEFAULT_SITE (now a real Sites value)
  and types defaultAuth as AuthOptionsWithDefaults so test fixtures match
  the tightened contract.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,7 +170,7 @@ The Datadog site to use APIs from, and which Datadog site telemetry metrics and 
 - `'ap1.datadoghq.com'`
 - `'ap2.datadoghq.com'`
 
-An unsupported value (passed via configuration or the `DATADOG_SITE` / `DD_SITE` environment variable) will throw at plugin initialization with a link back to the site list.
+An unsupported value (passed via configuration or the `DATADOG_SITE` / `DD_SITE` environment variable) will throw at plugin initialization.
 
 > [!NOTE]
 > The `DATADOG_SITE` environment variable takes priority over this configuration.
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Follow the specific documentation for each bundler:
     auth?: {
         apiKey?: string;
         appKey?: string;
-        site?: Sites;
+        site?: Site;
     };
     customPlugins?: (arg: GetPluginsArg) => UnpluginPlugin[];
     enableGit?: boolean;
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ Follow the specific documentation for each bundler:
     auth?: {
         apiKey?: string;
         appKey?: string;
-        site?: string;
+        site?: Sites;
     };
     customPlugins?: (arg: GetPluginsArg) => UnpluginPlugin[];
     enableGit?: boolean;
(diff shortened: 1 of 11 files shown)


oliverli:
* **#372** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/372?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/core/src/constants.ts
```diff
@@ -8,6 +8,20 @@ export const INJECTED_FILE_RX = new RegExp(INJECTED_FILE);
...
+export const SITES = [
+    'datadoghq.com',
+    'us3.datadoghq.com',
+    'us5.datadoghq.com',
+    'datadoghq.eu',
+    'ddog-gov.com',
+    'us2.ddog-gov.com',
+    'ap1.datadoghq.com',
+    'ap2.datadoghq.com',
+    'datad0g.com',
```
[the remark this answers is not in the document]
oliverli: Intentional, that's staging

packages/core/src/types.ts
```diff
@@ -244,10 +244,11 @@ export type GetWrappedPlugins = (arg: GetPluginsArg) => (PluginOptions | CustomP
 
 export type LogLevel = 'debug' | 'info' | 'warn' | 'error' | 'none';
 
+export type Sites = (typeof SITES)[number];
```
yoannmoinet: This should be singular.
```suggestion
export type Site = (typeof SITES)[number];
```

packages/factory/src/validate.ts
```diff
@@ -26,15 +47,21 @@ const validateMetadata = (metadata: BuildMetadata | undefined): string[] => {
...
     }
 
+    // DATADOG_SITE env var takes precedence over configuration.
+    // envSite is re-checked with isSite so TS narrows it from string to Sites.
+    const validatedEnvSite = envSite !== undefined && isSite(envSite) ? envSite : undefined;
     const auth: AuthOptionsWithDefaults = {
-        // DATADOG_SITE env var takes precedence over configuration
-        site: getDDEnvValue('SITE') || options.auth?.site || 'datadoghq.com',
+        site: validatedEnvSite ?? options.auth?.site ?? 'datadoghq.com',
     };
```
yoannmoinet: I feel like this could be greatly simplified and more readable by extracting the logic around `site` computation in its own helper, and done in a different order:
1. Define the `site` value from either env, options or default.
2. Validate it once.

No need to differentiate where the value is coming from in the error.

To also note that right now we're getting notified (and blocked) even if the incorrect value isn't used (from `auth.site`) which could be pretty misleading (eg. incorrect `auth.site` but correct `DD_SITE`).

packages/tests/src/_jest/helpers/mocks.ts
```diff
@@ -47,7 +49,7 @@ import path from 'path';
 
 import { getTempWorkingDir } from './env';
 
-export const FAKE_SITE = 'example.com';
+export const DEFAULT_SITE: Sites = 'datadoghq.com';
```
yoannmoinet: This should be a `core` constant, also used in the `validate`.

README.md
```diff
@@ -159,11 +159,18 @@ In order to interact with Datadog, you have to use [your own Application Key](ht
...
+- `'us3.datadoghq.com'`
+- `'us5.datadoghq.com'`
+- `'datadoghq.eu'`
+- `'ddog-gov.com'`
+- `'us2.ddog-gov.com'`
+- `'ap1.datadoghq.com'`
+- `'ap2.datadoghq.com'`
 
-This configuration controls which Datadog site telemetry metrics and error tracking sourcemaps are sent to.
+An unsupported value (passed via configuration or the `DATADOG_SITE` / `DD_SITE` environment variable) will throw at plugin initialization with a link back to the site list.
```
yoannmoinet: This seems weirdly specific...
```suggestion
An unsupported value (passed via configuration or the `DATADOG_SITE` / `DD_SITE` environment variable) will throw at plugin initialization.
```

yoannmoinet (approved):

PR #376, opened 2026-05-16
[APPS][Connections Part 13] Resolve static string values

sdkennedy2:
## Motivation

Imported backend connection IDs need two reusable analysis layers: first resolve an identifier to its static definition, then evaluate the resolved expression as a string in the module where that definition lives. The previous PR added the definition resolver; this PR adds the domain-neutral string value layer that connection ID extraction can consume next.

## Changes

Adds `resolveStaticStringValue`, a generic AST utility that resolves supported ESTree expressions to concrete string values across collected `ParsedModuleRecord` values:

```ts
resolveStaticStringValue(modules, moduleId, expression)
```

Example:

```ts
// actions.backend.ts
import { HTTP_CONNECTION_ID } from './connections';

request({
    connectionId: HTTP_CONNECTION_ID,
    inputs: {},
});

// connections.ts
const SOURCE_CONNECTION_ID = 'conn-http';
export const HTTP_CONNECTION_ID = SOURCE_CONNECTION_ID;
```

When called with the `HTTP_CONNECTION_ID` expression from `actions.backend.ts`, the resolver follows the identifier to the exported const in `connections.ts`, then evaluates that initializer in the definition module context:

```ts
resolveStaticStringValue(
    modules,
    '/project/src/backend/actions.backend.ts',
    HTTP_CONNECTION_ID,
);
// => { kind: 'resolved', value: 'conn-http' }
```

Static string resolution treats the definition resolver as the identifier lookup layer. Once the definition resolver returns:

```ts
export const HTTP_CONNECTION_ID = SOURCE_CONNECTION_ID;
```

the string resolver evaluates `SOURCE_CONNECTION_ID`, calls the static definition resolver again in the `connections.ts` module context to resolve that identifier, and finally evaluates:

```ts
const SOURCE_CONNECTION_ID = 'conn-http';
```

The resolver supports inline string literals, static template literals, same-module and imported const chains, and static object member paths such as `CONNECTIONS.HTTP.PROD`. It delegates identifier lookup to `resolveStaticDefinitionForIdentifier`, then evaluates the returned const expression in the definition module context.

Unsupported cases return structured results instead of throwing. Dynamic templates, non-string literals, computed or optional member reads, object spreads, computed keys, accessors, duplicate or missing properties, non-object member intermediates, const cycles, imported object mutations, and static-definition failures all produce explicit unsupported reasons with human-readable messages.

## QA Instructions

[...]

PR #375, opened 2026-05-16
[APPS][Connections Part 12] Use static definitions for imported connection IDs

sdkennedy2:
## Motivation

This is the next PR in the backend connection ID analysis stack. The previous PR adds a domain-neutral static definition resolver that can follow identifier references through named imports, export aliases, re-exports, and star exports. Connection ID extraction can now reuse that resolver instead of maintaining its own import/export traversal.

## Changes

Wires module-graph connection ID extraction into the static definition resolver. `extractConnectionIds(...)` now requires `ParsedModuleRecord` and module-map context so `connection-id-values.ts` can resolve the actual `connectionId` identifier, or the root identifier of a member expression, through `resolveStaticDefinitionForIdentifier(...)` and evaluate the returned const binding in the module where that definition lives.

Supported cases:

- Imported string constants
- Imported static template literals without interpolation
- Imported const-to-const chains across module contexts
- Imported object roots such as `CONNECTIONS.HTTP`
- Imported nested object member paths such as `CONNECTIONS.HTTP.PROD`
- Local export aliases
- Named re-export aliases
- Local import/export relays
- Unambiguous star exports
- Existing same-module inline, const, and object connection IDs

Unsupported cases continue to fail closed with connection ID errors:

- Missing exports
- Ambiguous star exports
- Import/export cycles
- Default imports
- Namespace imports
- Mutable bindings
- Reassigned bindings
- Unsupported top-level declarations and binding patterns
- Dynamic object shapes already rejected by the object resolver

This removes the older optional same-module binding-map path from connection ID value resolution. Same-module identifiers now use the same static definition resolver as imported identifiers.

## QA Instructions

No manual QA needed. This is an internal AST-analysis change covered by unit tests and does not change manifest wiring outside the backend connection ID extractor.

## Blast Radius

This affects High Code Apps backend connection ID extraction in `@dd/apps-plugin` when analyzing reachable app-local backend modules. It expands supported static import/re-export shapes while preserving fail-closed behavior for unsupported module graph and value shapes.

## Documentation

-   [How to reliably make frontend production changes](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/2286781319/Frontend+Production+Changesp)
-   [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)
-   [Deploy steps](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/1457750032/)

PR #374, opened 2026-05-15
[APPS][Connections Part 11] Resolve static definitions through module graph

sdkennedy2:
## Motivation

This is the next PR in the backend connection ID analysis stack. The previous PR enriches parsed backend module records with domain-neutral import, export, star export, and top-level binding facts. The imported connection ID integration needs a reusable way to resolve those facts through the backend module graph without embedding import/export traversal inside connection ID-specific code.

## Changes

Adds a domain-neutral static definition resolver for parsed backend module records. Given an ESTree identifier reference in a module, callers can resolve it to a concrete top-level const binding, or to a structured unsupported result with a stable reason code.

Example:

```ts
// actions.backend.ts
import { HTTP_CONNECTION_ID } from './connections';

request({
    connectionId: HTTP_CONNECTION_ID,
    inputs: {},
});

// connections.ts
const SOURCE_CONNECTION_ID = 'conn-http';
export const HTTP_CONNECTION_ID = SOURCE_CONNECTION_ID;
```

For the `HTTP_CONNECTION_ID` identifier in `actions.backend.ts`, the resolver follows the named import into `connections.ts` and returns the exported top-level const binding:

```ts
export const HTTP_CONNECTION_ID = SOURCE_CONNECTION_ID;
```

It does not evaluate the initializer into a string. That value-level evaluation is handled by the follow-up static string resolver PR.

The resolver is demand-driven: it follows named imports, local export aliases, named re-exports, local import/export relays, and unambiguous star exports for one requested symbol at a time.

Supported cases:

- Same-module identifier references to top-level const bindings
- Named imports through source module exports
- Local export aliases
- Named re-export aliases
- Local import/export relays
- Unambiguous star exports
- Explicit exports taking precedence over star exports

Unsupported cases return structured reason codes instead of throwing connection ID-specific errors.

Unsupported cases:

- Missing module records
- Missing exports
- Ambiguous star exports
- Import/export cycles
- Default imports
- Default exports and default re-exports
- Namespace imports
- Missing static bindings
- Mutable bindings
- Unsupported top-level declarations and binding patterns
- Unresolved identifiers and declaration identifiers

The result includes hop metadata so the follow-up connection ID PR can translate generic resolver failures into existing fail-closed diagnostics without recomputing graph traversal.

## QA Instructions

No manual QA needed. This is an internal AST-analysis helper with focused unit coverage and no manifest behavior changes in this PR.

## Blast Radius

[...]

PR #373, opened 2026-05-15
[APPS][Connections Part 9] Refactor action catalog module scope boundary

sdkennedy2:
## Motivation

This is the first slice split out of the previous all-in-one module graph PR. Before enriching module graph records, the shared eslint-scope analysis needs to live outside the action-catalog-specific call-site logic so module graph parsing can reuse identifier resolution without depending on connection ID or action catalog concepts.

## Changes

This introduces a domain-neutral `ModuleScopeAnalysis` helper for parsed backend ES modules. It owns the eslint-scope manager, reference lookup by identifier node, and module-scope variable lookup.

Action catalog parsing now layers `ActionCatalogScopeAnalysis` on top of `ModuleScopeAnalysis`:

```ts
export interface ActionCatalogScopeAnalysis {
    moduleScope: ModuleScopeAnalysis;
    actionFunctions: Set<eslintScope.Variable>;
    actionNamespaces: Set<eslintScope.Variable>;
}
```

Connection ID value extraction now receives the generic module scope analysis directly. That keeps identifier-to-definition resolution shared while leaving action-catalog call classification and connection ID value resolution as separate domain concerns.

## QA Instructions

No manual QA; this is an internal AST parsing refactor covered by CI.

## Blast Radius

This only affects backend apps AST parsing used to discover action-catalog connection IDs. It should be behavior-preserving: the same call-site and same-module connection ID extraction tests continue to cover the changed boundary.

## Documentation

-   [How to reliably make frontend production changes](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/2286781319/Frontend+Production+Changesp)
-   [PR Description Guidelines](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/3120334387/PR+Checklist+and+Guidelines+for+Web+UI)
-   [Deploy steps](https://datadoghq.atlassian.net/wiki/spaces/FRON/pages/1457750032/)

PR #377, opened 2026-05-16
[APPS][Connections Part 14] Use static string values for imported connection IDs

sdkennedy2:
## Motivation

Backend connection ID extraction already walks reachable app-local modules, but it only supported inline and same-module values. Real apps often keep connection IDs in shared modules and import them into backend helpers. This PR wires the generic static string resolver into module-graph connection ID extraction so those imported values can be analyzed conservatively.

## Changes

Module-graph extraction now uses each reachable `ParsedModuleRecord` directly instead of reparsing through the same-file-only extraction path. For each action-catalog call, the connection ID adapter passes the full `connectionId` expression to:

```ts
resolveStaticStringValue(modules, record.id, connectionIdExpression)
```

Resolved strings are added to the conservative backend-file allowlist. Unsupported results fail closed using the existing connection ID error path while preserving the generic unsupported reason and message for diagnostics.

This broadens supported connection ID shapes to imported constants, imported static templates, imported const chains, imported object roots, nested static object paths, local export aliases, named re-exports, import/export relays, and unambiguous star exports. Default imports, namespace imports, ambiguous star exports, missing exports, mutable or reassigned bindings, cycles, and dynamic expressions remain unsupported and fail closed.

## QA Instructions

Review `extract-connection-ids-from-module-graph.test.ts` for the imported constant, const-chain, object-path, re-export, star-export, and unsupported-case coverage. The stack remains conservative: unsupported imported or dynamic forms throw instead of silently omitting a possible connection ID.

## Blast Radius

This affects backend manifest connection ID extraction for reachable app-local backend modules. It does not change module collection, graph walking, upload plumbing, dev middleware, or the manifest schema.

## Documentation


sdkennedy2:
* **#377** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/377?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/377?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#376** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/376?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#374** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/374?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#375](https://github.com/DataDog/build-plugins/pull/375) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/375?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#371** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/371?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#373** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/373?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

sarenji (approved):
LGTM. One observation: Seems like the tests are quite similar across PRs. I wonder if we only need to test the public interface (as you do in this PR), rather than the implementation details. For example if a change goes wrong, would we have multiple tests fail across the stack? But you know best whether the more "implementation-y"/unit-y tests are a useful signal to have as well.

sdkennedy2:
Thanks, that makes sense. I tried to move this PR in that direction by deleting the lower-level extractConnectionIds/connection-id-values tests and covering the behavior through extractConnectionIdsFromModuleGraph instead. I kept the import/re-export and fail-closed cases there because regressions could silently produce a bad connection allowlist, but agreed we should avoid duplicating the same lower-level cases across the stack.

PR #379, opened 2026-05-18
Read backend-function args from $.backendFunctionArgs to fix code injection

oliverli:
## Summary

Removes a code-injection vulnerability in the generated production script for `@dd/apps-plugin` backend functions.

The script generated by `virtual-entry.ts` read args via textual substitution into a single-quoted JS string literal:

```js
const args = JSON.parse('${backendFunctionArgs}' || '[]');
```

When the host runtime (web-ui) substituted `${backendFunctionArgs}` with the JSON of the iframe's args, any string arg containing `'` broke out of the literal. `JSON.stringify` does not escape single quotes, so this manifested two ways:

1. **Functional bug:** common English contractions (`don't`, `won't`) sent to any backend function produced `Parsing error: Expected ',', got 's' at <line>:<col>` — the host's parser choking on the closing quote and the next token of the JSON.
2. **Code injection:** a user-controlled string like `'); await $.Actions.<conn>.exfiltrate(...); //` escaped the literal and ran in the backend executor's privileged context (datastore access, connection credentials).

This change moves args off the textual-substitution path entirely. The generated script now reads:

```js
const args = $.backendFunctionArgs ?? [];
```

The companion PR in web-ui (DataDog/web-ui#307916) places args into `inputs.context.backendFunctionArgs` as a structured JSON value before sending the query to the API. Values round-trip losslessly for any Unicode content.

Dev and production codegen are unified — `generateDevVirtualEntryContent` now delegates to `generateVirtualEntryContent` so both paths share the same runtime contract. The dev-server (`/__dd/executeAction`) threads args through to `inputs.context.backendFunctionArgs` in the preview-async POST instead of inlining them into the bundle.

## Changes

- `packages/plugins/apps/src/backend/virtual-entry.ts` — production codegen reads `$.backendFunctionArgs`; dev codegen delegates to production (drops the `args` parameter).
- `packages/plugins/apps/src/vite/dev-server.ts` — `bundleBackendFunction` no longer takes args; `executeScriptViaDatadog` sets `spec.inputs.context.backendFunctionArgs`.
- `packages/plugins/apps/src/backend/virtual-entry.test.ts` — updated assertions to require `$.backendFunctionArgs`; new round-trip test evaluates the generated `main` against adversarial inputs (`don't break`, `'); alert(1); //`, mixed quotes/backslashes, template-literal syntax, newlines/tabs, emoji).
- `packages/plugins/apps/src/vite/dev-server.test.ts` — captures the preview-async body and asserts args appear in `inputs.context` with `template_params` empty; new single-quote round-trip test.

## Coordination

Older apps already deployed against the previous build-plugins release read via `template_params.backendFunctionArgs` and won't pick up this fix until rebuilt + redeployed. The matching web-ui PR removes the `template_params.backendFunctionArgs` write entirely (rather than keeping it during a transition) so no exploitable path remains — older apps will fail loudly until rebuilt. App authors will need to be notified to redeploy.

## Test plan

- [x] `yarn lint` on changed files — clean
- [x] `yarn cli typecheck-workspaces` — clean
- [x] `yarn cli integrity` — clean
- [x] `yarn test:unit packages/plugins/apps` — 234 tests pass (1 pre-existing stale `.d.ts` artifact in `dist/`, unrelated)
- [x] Manual: scaffold an App, build with this branch, deploy alongside web-ui#307916, call a backend function with `args: ["don't break"]`, confirm the value round-trips into the datastore

🤖 Generated with Claude Opus 4.7

### DIFF
--- a/packages/plugins/apps/src/backend/virtual-entry.test.ts
+++ b/packages/plugins/apps/src/backend/virtual-entry.test.ts
@@ -47,14 +47,20 @@ describe('Backend Functions - generateVirtualEntryContent', () => {
             expect(result).toContain('globalThis.$ = $');
         });
 
-        test('Should include backendFunctionArgs template expression', () => {
+        test('Should read args from $.backendFunctionArgs (no source-text substitution)', () => {
             const result = generateVirtualEntryContent(
                 'myHandler',
                 '/src/handler.ts',
                 PROJECT_ROOT,
             );
+            expect(result).toContain('const args = $.backendFunctionArgs ?? [];');
+            // The previous design textually substituted args into a single-quoted
+            // string literal, which allowed `'` in user input to break out and
+            // inject arbitrary JS. The generated script must never wrap a host
+            // template expression inside a quoted string.
             // eslint-disable-next-line no-template-curly-in-string
-            expect(result).toContain("JSON.parse('${backendFunctionArgs}'");
+            expect(result).not.toContain('${backendFunctionArgs}');
+            expect(result).not.toContain('JSON.parse(');
         });
 
         test('Should call the function with spread args', () => {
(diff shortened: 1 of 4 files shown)


oliverli:
* **#379** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/379?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/379?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

oliverli:
@cursor review
@codex review

sdkennedy2 (approved):

PR #381, opened 2026-05-19
Fix corrupted ESM bundles in published packages

oliverli:
### What and why?

The published `.mjs` files for all five `@datadog/*-plugin` packages have been shipping broken since v3.1.5. They fail `node --check` and crash any ESM consumer (Vite picks `.mjs` via `exports.import`) with `SyntaxError`. The shape of the corruption: the `// -- Shims --` block from `@rollup/plugin-esm-shim` is injected at a wrong byte offset — in vite-plugin@3.1.7 it lands inside the `SITES` array, splitting the string literal `"ap2.datadoghq.com"` into `"ap2.datado` + shim block + `ghq.com"`.

Root cause is upstream in `@rollup/plugin-esm-shim@0.1.7`: its `matchAllPolyfill` calls `String.prototype.match(regex)` (which returns matched strings without positions) and then synthesizes match indices by accumulating match lengths — only correct if every match is contiguous from offset 0. In the minified single-line bundle, the regex picks up `import ... from "..."` substrings inside string literals as if they were real imports, and the polyfill's sum-of-lengths math overshoots well past the actual import block.

### How?

Two coupled changes:

- **Bump `@rollup/plugin-esm-shim` 0.1.7 → 0.1.8** across the six `package.json` files that pin it. Upstream replaced the broken polyfill with `mlly`'s AST-aware `findStaticImports`, so insertion offsets are now real byte positions.
- **Reorder rollup output plugins** in `packages/tools/src/rollupConfig.mjs` so `esmShim()` runs before `terser()`. Defensive: even with 0.1.8, running the shim plugin on un-minified output is conceptually correct (shim-then-minify) and avoids feeding minified single-line code to any future regex-based scan. Side benefit: none of the five bundles actually use `__dirname`/`__filename`/`require(`, so with the new order the shim block is correctly omitted entirely.

### Verification

- `node --check` now passes on all five `packages/published/*/dist/src/index.mjs` outputs (was failing on vite-plugin before this change).
- `yarn test:unit packages/tools/src/rollupConfig.test.ts` — 16 tests pass across all five bundlers.
- `yarn lint` and `yarn workspace @dd/tools typecheck` on the changed file: clean.

### Follow-up

A patch release of all five published packages will be needed to ship clean `.mjs` artifacts to consumers. Existing 3.1.5–3.1.7 releases on npm remain broken.

🤖 Generated with Claude Opus 4.7 (1M context)


oliverli:
* **#381** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/381?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/381?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

yoannmoinet (approved):
That's a lot of dependencies for a patch bump.
Other than that, looks good 👍

PR #392, opened 2026-06-02
Add sourcemap upload failure metrics

watson:
### What and why?

Add opt-in Datadog metrics for Error Tracking sourcemap upload retries and final upload failures. These metrics make transient sourcemap intake issues and exhausted retry failures observable without parsing CI logs.

The metrics are emitted only when the existing `metrics` plugin is enabled, so customers using Error Tracking without build metrics keep the same default experience.

### How?

Sourcemap uploads record retry and failure counts in memory during each upload phase, then add those counts to a shared internal metrics collector on the global context.

The metrics plugin owns the actual metrics pipeline: it includes collected upload metrics in the existing `metrics: {}` payload, applies the normal filtering, prefixing, tags, custom metrics hook, and Datadog send behavior, and awaits the send. It also flushes any metrics collected late at true end so fallback sourcemap uploads keep the same awaited behavior.

Metrics added before the metrics plugin prefix is applied:

- `sourcemaps.upload.retry`
- `sourcemaps.upload.failure`

Tags include bundler, plugin version, service, site, status code, error type, and CI job context when available.

### Testing

- `yarn workspace @dd/metrics-plugin typecheck`
- `yarn workspace @dd/error-tracking-plugin typecheck`
- `yarn test:unit packages/plugins/error-tracking/src/sourcemaps/sender.test.ts packages/plugins/error-tracking/src/index.test.ts packages/plugins/metrics/src/index.test.ts`


watson:
* **#392** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/392?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/392?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
```diff
@@ -0,0 +1,133 @@
...
+
+const getBaseMetricTags = (options: SourcemapsOptionsWithDefaults, context: UploadContext) => [
+    `bundler:${context.bundlerName}`,
+    `plugin_version:${context.version}`,
+    `service:${options.service}`,
+    `site:${context.site}`,
+    ...(process.env.CI_JOB_NAME ? [`jobname:${normalizeTagValue(process.env.CI_JOB_NAME)}`] : []),
+    ...(process.env.BRANCH_TYPE
+        ? [`branchtype:${normalizeTagValue(process.env.BRANCH_TYPE)}`]
+        : []),
```
yoannmoinet: I feel like these should be default for all metrics we send through the metrics plugin.
Or at least, it feels weird to me to have different metrics definition between base metrics and these new metrics.
It should be made through an unified pipeline.

Also, `CI_JOB_NAME` and `BRANCH_TYPE` are pretty custom to be included here.
Aren't all of these already handled on the user's side (web-ui)?

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
yoannmoinet, line 1: A bunch of these helpers are only used for sourcemaps metrics but are made in a generic way, as-in are not really sourcemaps specific.

They should instead be part of the main metrics pipeline we already have.
So all the metrics we send can benefit from it.

packages/plugins/metrics/src/index.ts
```diff
@@ -69,7 +104,13 @@ export const getPlugins: GetPlugins = ({ options, context }) => {
...
         timeLoader.end();
 
-        const allMetrics = new Set([...universalMetrics, ...pluginMetrics, ...loaderMetrics]);
+        const allMetrics = new Set([
+            ...universalMetrics,
+            ...pluginMetrics,
+            ...loaderMetrics,
+            ...stores.metrics,
+        ]);
+        stores.metrics.clear();
```
yoannmoinet: I don't think we should change this part, and only rely on the "flush" mechanism for these metrics added at build time.

packages/plugins/metrics/src/index.ts
```diff
@@ -129,6 +170,9 @@ export const getPlugins: GetPlugins = ({ options, context }) => {
                 await computeMetrics();
             }
         },
+        async asyncTrueEnd() {
+            await sendCollectedMetrics();
+        },
```
yoannmoinet: That's great to have this flush mechanism here.
But it reveals a design flaw in how we run the `asyncTrueEnd` hook.

Right now, the sourcemaps are also uploaded part of the `asyncTrueEnd` hook, so we may end up with a race condition as there is no enforced order during the `trueEnd` hooks.

Maybe we could introduce a new `flush` hook that would execute after the `asyncTrueEnd`/`syncTrueEnd` hooks to ensure there is no race condition for this specific use case.

It conflicts with the `trueEnd` semantics though 😔 but I don't really have a better idea for this.

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
watson: Confirmed: `jobname` and `branchtype` are already provided by web-ui via `defaultMetricTags` (`jobname` from `CI_JOB_NAME`, `branchtype` from git branch info), passed through the plugin's `tags` option.

I've dropped the custom `CI_JOB_NAME`/`BRANCH_TYPE` handling. The sourcemap metrics now go through the unified pipeline, so they pick up those configured tags automatically. 

`bundler`/`plugin_version`/`site` aren't set by web-ui, so I kept those as defaults but applied them to all metrics rather than just sourcemaps.

packages/plugins/metrics/src/index.ts
watson: Fixed 👍 I left the naming issue to a future PR as a TODO comment.

packages/plugins/metrics/src/index.ts
watson: Fixed 👍

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
watson: Fixed 👍

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
```diff
@@ -0,0 +1,114 @@
...
+    return match ? `status_code:${match[1]}` : 'status_code:unknown';
+};
+
+const getErrorTypeTag = (error: Error): string => {
+    const statusCodeTag = getStatusCodeTag(error);
+    if (statusCodeTag !== 'status_code:unknown') {
+        return `error_type:http_${statusCodeTag.replace('status_code:', '')}`;
+    }
+
+    return `error_type:${error.name || 'unknown'}`;
```
yoannmoinet: nit: this could use `normalizeTagValue`.
```suggestion
    return `error_type:${normalizeTagValue(error.name || 'unknown')}`;
```
(to be imported)

packages/plugins/error-tracking/src/sourcemaps/sender.test.ts
```diff
@@ -190,7 +195,9 @@ describe('Error Tracking Plugin Sourcemaps', () => {
         });
 
         test('Should alert in case of errors', async () => {
-            doRequestMock.mockRejectedValue(new Error('Fake Error'));
+            doRequestMock
+                .mockRejectedValueOnce(new Error('Fake Error'))
+                .mockResolvedValueOnce(undefined);
```
yoannmoinet: nit: This is maybe a leftover, but IIUC, this is never reached. Especially if we do `expect(doRequestMock).toHaveBeenCalledTimes(1)`.

packages/plugins/error-tracking/src/sourcemaps/sender.test.ts
```diff
@@ -209,10 +216,13 @@ describe('Error Tracking Plugin Sourcemaps', () => {
...
             });
             expect(warnings).toHaveLength(0);
+            expect(doRequestMock).toHaveBeenCalledTimes(1);
         });
 
         test('Should throw in case of errors with bailOnError', async () => {
-            doRequestMock.mockRejectedValue(new Error('Fake Error'));
+            doRequestMock
+                .mockRejectedValueOnce(new Error('Fake Error'))
+                .mockResolvedValueOnce(undefined);
```
yoannmoinet: nit: same as previous, I don't think this second mock is reached. To be confirmed.

yoannmoinet (approved):
This is really good.
Thanks for sticking around with my _many_ requests, really appreciated.

Just left some nits, this could be shipped as-is.

packages/plugins/error-tracking/src/sourcemaps/upload-metrics.ts
watson: Good point. I couldn’t import `normalizeTagValue` directly from `@dd/metrics-plugin` because `@dd/error-tracking-plugin` doesn’t depend on it.

I moved the generic helper to `@dd/core/helpers/strings` and updated both metrics and sourcemap upload metrics to use it from there - I hope that's ok

packages/plugins/error-tracking/src/sourcemaps/sender.test.ts
watson: Ups, fixed 👍

packages/plugins/error-tracking/src/sourcemaps/sender.test.ts
watson: Ups, fixed 👍

PR #393, opened 2026-06-02
[APPS] Add OAuth auth for Apps

otorrillas:
## Summary

Depends on #421.

Adds OAuth support for Datadog Apps while keeping the optional dependency integrity support in its own PR:

- adds the core OAuth PKCE token flow and optional keychain cache
- adds `oauth-request.ts`, which owns OAuth token resolution and `doOAuthRequest`
- extends `doRequest` to accept request-ready auth via API/app keys or an already resolved OAuth access token
- adds Apps-scoped auth selection via `apps.authOverrides.method` and `DD_APPS_AUTH_METHOD` / `DATADOG_APPS_AUTH_METHOD`, defaulting to API-key auth
- wires OAuth auth into Apps uploads and the Apps Vite dev-server execution path
- keeps `oauth4webapi` as a regular dependency and `@napi-rs/keyring` as an optional dependency

## Validation

- `yarn workspace @dd/tests test:unit packages/core/src/helpers/request.test.ts packages/core/src/helpers/oauth-request.test.ts packages/plugins/error-tracking/src/sourcemaps/sender.test.ts packages/plugins/apps/src/auth.test.ts packages/plugins/apps/src/upload.test.ts packages/plugins/apps/src/index.test.ts packages/plugins/apps/src/validate.test.ts packages/plugins/apps/src/vite/dev-server.test.ts packages/plugins/apps/src/vite/index.test.ts`
- `yarn cli integrity`
- `git diff --check`


packages/core/src/helpers/oauth-request.ts
yoannmoinet, line 1: This should be factorised around the `doRequest` helper as much as possible.

What if another plugin needs to do oauth, will they have to re-implement most of this? (honest question, not rhetorical 😄)

packages/core/src/types.ts
```diff
@@ -245,9 +245,13 @@ export type GetWrappedPlugins = (arg: GetPluginsArg) => (PluginOptions | CustomP
 export type LogLevel = 'debug' | 'info' | 'warn' | 'error' | 'none';
 
 export type Site = (typeof SITES)[number];
+export type AuthMethod = 'apiKey' | 'oauth';
+
 export type AuthOptions = {
     apiKey?: string;
     appKey?: string;
+    accessToken?: string;
```
yoannmoinet: Why's this user facing? Not sure I understand the need for it, from the user's perspective.

packages/core/src/types.ts
```diff
@@ -283,7 +287,7 @@ export type PluginName = `datadog-${Lowercase<string>}-plugin`;
 type Data = { data?: BodyInit; headers?: Record<string, string> };
 export type RequestOpts = {
     url: string;
-    auth?: Pick<AuthOptions, 'apiKey' | 'appKey'>;
+    auth?: Pick<AuthOptions, 'apiKey' | 'appKey' | 'accessToken'>;
```
yoannmoinet: I believe this should be more of the form:
```suggestion
    auth?: Assign<Pick<AuthOptions, 'apiKey' | 'appKey'>, { accessToken: string }>;
```

So `accessToken` isn't leaked into user facing API (unless we actually need it there).

packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -27,7 +27,7 @@ type BundleFn = (func: BackendFunction) => Promise<BundleResult>;
 
 const DEV_VIRTUAL_PREFIX = 'virtual:dd-backend-dev:';
 
-type AuthConfig = Required<AuthOptionsWithDefaults>;
+type AuthConfig = Required<Pick<AuthOptionsWithDefaults, 'apiKey' | 'appKey' | 'site'>>;
```
yoannmoinet: This won't be necessary with my previous comment.

yoannmoinet (commented):
Doing a quick review first, to tackle my biggest comments 😃

packages/core/src/helpers/request.ts
```diff
@@ -117,12 +133,16 @@ export const doRequest = <T>(opts: RequestOpts): Promise<T> => {
             };
 
             // Do auth if present.
-            if (auth?.apiKey) {
-                requestHeaders['DD-API-KEY'] = auth.apiKey;
-            }
+            if (auth?.authMethod === 'oauth') {
+                requestHeaders.Authorization = `Bearer ${accessToken}`;
```
sdkennedy2: Is it worth us validating that accessToken exists?

packages/plugins/apps/src/vite/handle-upload.ts
```diff
@@ -139,11 +139,11 @@ Either:
         archiveDir = path.dirname(archive.archivePath);
 
         const uploadTimer = log.time('upload assets');
+        const requestAuth = getRequestAuth(options.method, auth);
```
sdkennedy2: Having `getRequestAuth` live in upload seems odd since it will probably be used across the entire apps plugin

packages/plugins/apps/src/index.test.ts
```diff
@@ -230,6 +231,93 @@ describe('Apps Plugin - getPlugins', () => {
         expect(fsHelpers.rm).toHaveBeenCalledWith(expect.stringContaining('dd-apps-manifest-'));
     });
 
+    test('Should pass the OAuth method through to the uploader when configured', async () => {
```
sdkennedy2: This kind of feels like we are reaching into the implementation details a bit.

packages/plugins/apps/src/types.ts
```diff
@@ -28,4 +33,6 @@ export type AppsManifest = {
 export type AppsOptionsWithDefaults = Omit<
     WithRequired<AppsOptions, 'include' | 'dryRun'>,
     'enable'
->;
+> & {
+    method: AuthMethod;
+};
```
sdkennedy2: Why is `AppsOptionsWithDefaults` such a different shape?

packages/plugins/apps/src/upload.ts
```diff
@@ -113,9 +125,20 @@ Would have uploaded ${summary}`,
         return { errors, warnings };
     }
 
+    const requestAuth = context.auth;
```
sdkennedy2: Should this go above `if (!context.identifier) {` to maintain similar ordering.

packages/plugins/apps/src/validate.ts
```diff
@@ -3,15 +3,34 @@
...
+    }
+
+    throw new Error(`apps.authOverrides.method must be one of: ${AUTH_METHODS.join(', ')}`);
+};
+
 export const validateOptions = (options: Options): AppsOptionsWithDefaults => {
     const resolvedOptions = (options[CONFIG_KEY] || {}) as AppsOptions;
+    const method =
+        resolveAuthMethod(getDDEnvValue('AUTH_METHOD')) ||
+        resolveAuthMethod(resolvedOptions.authOverrides?.method) ||
```
sdkennedy2: Wouldn't this make more sense?
```suggestion
        resolveAuthMethod(getDDEnvValue('AUTH_METHOD') || resolvedOptions.authOverrides?.method) ||
```

packages/plugins/apps/README.md
```diff
@@ -18,19 +18,29 @@ A plugin to upload assets to Datadog's storage
...
 
 ## Configuration
 
 ```ts
+auth?: {
+    apiKey?: string;
+    appKey?: string;
+    site?: string;
+}
+
```
sdkennedy2: Should this be included?

packages/core/src/helpers/oauth.ts
```diff
@@ -0,0 +1,619 @@
...
+              : { command: 'xdg-open', args: [url] };
+
+    try {
+        const child = spawn(opener.command, opener.args, {
+            detached: true,
+            stdio: 'ignore',
+        });
+        child.unref();
+    } catch {
+        // Logging the URL is the reliable fallback.
```
sdkennedy2: Should we handle this somehow? Perhaps log?

packages/core/src/helpers/request.ts
```diff
@@ -94,16 +96,30 @@ export const createRequestData = async (options: {
...
 
-    return retry(async (bail: (e: Error) => void, attempt: number) => {
+    // Resolve the OAuth token once, before the retry loop, so the interactive
+    // browser authorization never runs per attempt.
+    let accessToken: string | undefined;
+    if (auth?.authMethod === 'oauth') {
+        if (!auth.site) {
+            throw new Error('OAuth authentication requires a site.');
+        }
+        if (!log) {
```
sdkennedy2: We have to do this because logs actually uses doRequest under the hood.

packages/published/esbuild-plugin/package.json
```diff
@@ -52,12 +52,14 @@
     "dependencies": {
         "@datadog/js-instrumentation-wasm": "1.0.8",
         "@jridgewell/remapping": "2.3.5",
+        "@napi-rs/keyring": "1.3.0",
```
sdkennedy2: If these are in the core package do we need these as dependencies within all the build specific plugins (esbuild-plugin, rollup-plugin, etc)

sdkennedy2: Yeah, I think we still need them here. The published packages bundle the internal `@dd/*` packages rather than depending on `@dd/core` at runtime, and the integrity dependency sync flattens external deps from internal workspaces into each published package. Since the OAuth helper in core dynamically loads `oauth4webapi` and `@napi-rs/keyring`, the published packages need to declare them so they’re installed for consumers.

packages/core/src/helpers/oauth.ts
```diff
@@ -0,0 +1,622 @@
...
+
+type OAuthAuthorizationServer = import('oauth4webapi').AuthorizationServer;
+type OAuthClient = import('oauth4webapi').Client;
+type OAuthModule = typeof import('oauth4webapi');
+type OAuthTokenEndpointResponse = import('oauth4webapi').TokenEndpointResponse;
+type KeyringModule = typeof import('@napi-rs/keyring');
+
+// Load through a variable specifier so bundlers leave these as runtime requires
+// instead of inlining them — notably the native `@napi-rs/keyring` binary. Node
+// caches modules by specifier, so repeated calls reuse the same instance.
```
sdkennedy2: Does this make sense?

packages/plugins/apps/src/upload.ts
```diff
@@ -113,9 +108,19 @@ Would have uploaded ${summary}`,
         return { errors, warnings };
     }
 
+    if (!requestAuth) {
```
sdkennedy2: It looks like our auth check got moved below ` if (!context.identifier) {`. Is this correct?

sdkennedy2:
@codex review
@cursor review

packages/core/src/types.ts
```diff
@@ -280,10 +282,21 @@ export type OptionsWithDefaults = Assign<
...
 
+// Request-local auth. `accessToken` is never accepted here — the OAuth path
+// is selected with `{ authMethod: 'oauth', site }` and `doRequest` resolves and
+// caches the token internally.
+export type RequestAuthOptions = {
+    authMethod?: AuthMethod;
+    apiKey?: string;
+    appKey?: string;
+    site: string;
+};
```
yoannmoinet: Why is `site` now mandatory?

From the implementation it doesn't seem to be used all the time, but only for the OAuth flow.

The typing should reflect it.
Maybe something like (not tested):
```suggestion
export type RequestAuthOptions = {
    authMethod: 'apiKey';
    apiKey?: string;
    appKey?: string;
} | {
    authMethod: 'oauth';
    site: string;
};
```

Also don't we have a `Site` type for internal `site`?

packages/core/src/helpers/request.ts
```diff
@@ -94,16 +96,33 @@ export const createRequestData = async (options: {
...
         maxTimeout: opts.maxTimeout,
         minTimeout: opts.minTimeout,
     };
 
-    return retry(async (bail: (e: Error) => void, attempt: number) => {
+    // Resolve the OAuth token once, before the retry loop, so the interactive
+    // browser authorization never runs per attempt.
+    let accessToken: string | undefined;
+    if (auth?.authMethod === 'oauth') {
+        if (!auth.site) {
```
yoannmoinet: Don't we always fallback to a default? Shouldn't we?

packages/core/src/helpers/oauth.ts
```diff
@@ -0,0 +1,622 @@
...
+const loadOauth = () => import(OAUTH_PACKAGE_NAME) as Promise<OAuthModule>;
+const loadKeyring = () => import(KEYRING_PACKAGE_NAME) as Promise<KeyringModule>;
```
yoannmoinet: Why do we do this if the dependency is required in all the `package.json`?

packages/core/src/helpers/oauth-request.ts
yoannmoinet, line 1: These (test file too) need a codeowner.

packages/plugins/apps/src/upload.ts
yoannmoinet: Shouldn't this be reflected on the type instead?

packages/plugins/error-tracking/src/sourcemaps/sender.ts
```diff
@@ -124,7 +124,7 @@ export const upload = async (
             queue.add(async () => {
                 try {
                     await doRequest({
-                        auth: { apiKey: context.apiKey },
+                        auth: { apiKey: context.apiKey, site: context.site },
```
yoannmoinet: This should not be necessary.

packages/plugins/apps/src/validate.ts
```diff
@@ -3,18 +3,39 @@
...
+    }
+
+    throw new Error(`apps.authOverrides.method must be one of: ${AUTH_METHODS.join(', ')}`);
+};
+
 export const validateOptions = (options: Options): AppsOptionsWithDefaults => {
     const resolvedOptions = (options[CONFIG_KEY] || {}) as AppsOptions;
+    const method =
+        resolveAuthMethod(
+            getDDEnvValue('APPS_AUTH_METHOD') || resolvedOptions.authOverrides?.method,
```
yoannmoinet, line 29: Are we confident on the order of precedence here?
We do want the env overriding what's defined in the options.

packages/core/src/helpers/request.ts
```diff
@@ -94,16 +96,33 @@ export const createRequestData = async (options: {
...
+            throw new Error('OAuth authentication requires a site.');
+        }
+        if (!log) {
+            throw new Error('OAuth authentication requires a logger.');
```
yoannmoinet: Why?

packages/plugins/apps/README.md
yoannmoinet: This will break the root README once included in it.
Because we copy this `## Configuration` block into the global configuration block of the root.

packages/plugins/apps/README.md
```diff
@@ -66,6 +76,22 @@ Must be a boolean. Non-boolean values are coerced today but will be rejected in
 
 Additional glob patterns (relative to the project root) to include in the uploaded archive. The bundler output directory is always included.
 
+### apps.authOverrides.method
+
+> default: `apiKey`
+
+Authentication method for uploading app bundles. Scoped to the `apps` plugin rather than the shared `auth` config, because not every Datadog endpoint supports OAuth.
```
yoannmoinet: Not sure this is a necessary detail...
```suggestion
Authentication method for uploading app bundles.
```

packages/core/src/helpers/oauth-request.ts
```diff
@@ -0,0 +1,622 @@
...
+
+    return undefined;
+};
+
+const createOAuthCredentialEntry = async (
+    site: string,
+    options: Pick<OAuthConfig, 'authorizationUrl' | 'clientId' | 'tokenUrl'>,
+) => {
+    const { AsyncEntry } = await loadKeyring();
+    return new AsyncEntry(OAUTH_KEYCHAIN_SERVICE, getOAuthCredentialAccount(site, options));
```
yoannmoinet, line 408: What is keyring necessary for?
No alternative?

packages/published/esbuild-plugin/package.json
yoannmoinet: This is pretty bad.
We need to find a way to make this optional only.

packages/core/src/types.ts
sdkennedy2: I had made the changes locally but forgot to put up the changes end of day Friday 🤦. I agree. Good call 👍

packages/core/src/helpers/request.ts
```diff
@@ -93,8 +95,29 @@ export const createRequestData = async (options: {
 
 export const ERROR_CODES_NO_RETRY = [400, 401, 403, 404, 405, 409, 413];
 export const NB_RETRIES = 5;
+
+export type OAuthRequestOpts = Omit<RequestOpts, 'auth'> & {
+    site: Site;
+    log: Logger;
+};
+
+export const doOAuthRequest = async <T>({ site, log, ...opts }: OAuthRequestOpts): Promise<T> => {
```
sdkennedy2: I wonder if this should be in here or closer to `oauth.ts`

packages/core/src/helpers/request.ts
```diff
@@ -103,7 +126,7 @@ export const doRequest = <T>(opts: RequestOpts): Promise<T> => {
         minTimeout: opts.minTimeout,
     };
 
-    return retry(async (bail: (e: Error) => void, attempt: number) => {
```
sdkennedy2, line 106: What this change intentional?

packages/core/src/helpers/request.ts
```diff
@@ -117,12 +140,19 @@ export const doRequest = <T>(opts: RequestOpts): Promise<T> => {
...
 
             // Do auth if present.
-            if (auth?.apiKey) {
-                requestHeaders['DD-API-KEY'] = auth.apiKey;
-            }
+            if (auth && 'accessToken' in auth) {
+                if (!auth.accessToken) {
+                    throw new Error('OAuth authentication did not return an access token.');
+                }
+                requestHeaders.Authorization = `Bearer ${auth.accessToken}`;
```
sdkennedy2: Lets align this to how we do it for the `apiKey` and `appKey`. Lets just set the Authorization header if the accessToken is defined.

packages/core/src/types.ts
```diff
@@ -245,6 +245,8 @@ export type GetWrappedPlugins = (arg: GetPluginsArg) => (PluginOptions | CustomP
 export type LogLevel = 'debug' | 'info' | 'warn' | 'error' | 'none';
 
 export type Site = (typeof SITES)[number];
+export type AuthMethod = 'apiKey' | 'oauth';
```
sdkennedy2: Do we need this anymore?

packages/core/package.json
```diff
@@ -26,7 +26,11 @@
         "async-retry": "1.3.3",
         "chalk": "2.3.1",
         "glob": "11.1.0",
-        "json-stream-stringify": "3.1.6"
+        "json-stream-stringify": "3.1.6",
+        "oauth4webapi": "3.8.6"
```
sdkennedy2, line 30: Should this be an optional dependency as well?

packages/plugins/apps/src/auth.ts
```diff
@@ -0,0 +1,33 @@
...
+// Copyright 2019-Present Datadog, Inc.
+
+import { doOAuthRequest, doRequest } from '@dd/core/helpers/request';
+import type { AuthMethod, AuthOptionsWithDefaults, Logger, RequestOpts } from '@dd/core/types';
+
+export type AuthenticatedRequest = <T>(opts: Omit<RequestOpts, 'auth'>) => Promise<T>;
+
+// Build the authenticated request function from the resolved method + base credentials.
+// Returns undefined when API-key auth is selected but credentials are missing.
+export const getAuthenticatedRequest = (
```
sdkennedy2, line 27: The naming here seems odd. We call the sub-calls `doOAuthRequest` and `doRequest`. Should this function return a `doAuthenticatedRequest`?

packages/plugins/apps/src/auth.ts
```diff
@@ -0,0 +1,33 @@
+// Unless explicitly stated otherwise all files in this repository are licensed under the MIT License.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2019-Present Datadog, Inc.
+
+import { doOAuthRequest, doRequest } from '@dd/core/helpers/request';
+import type { AuthMethod, AuthOptionsWithDefaults, Logger, RequestOpts } from '@dd/core/types';
+
+export type AuthenticatedRequest = <T>(opts: Omit<RequestOpts, 'auth'>) => Promise<T>;
```
sdkennedy2: I wonder if `@dd/core/helpers/request` should return a `DoRequest` type or if we just do `typeof doRequest`.

packages/plugins/apps/src/upload.ts
```diff
@@ -113,9 +104,17 @@ Would have uploaded ${summary}`,
         return { errors, warnings };
     }
 
+    if (!request) {
```
sdkennedy2: Can we just make this required?

packages/plugins/apps/src/validate.ts
sdkennedy2: Yes this is consistent with other places in the codebase:
https://github.com/DataDog/build-plugins/blob/master/packages/factory/src/validate.ts#L69

packages/tools/src/commands/integrity/dependencies.ts
```diff
@@ -83,15 +84,36 @@ const getInternalDependencies = (
     return { errors, dependencies };
 };
 
+type DependencyType = 'dependencies' | 'optionalDependencies';
```
sdkennedy2, line 87: Would it be worth separating out optionalDependencies integrity support into a separate PR?

packages/core/src/helpers/request.ts
sdkennedy2: I kept `doOAuthRequest` in `request.ts` because it composes token resolution with request execution. `oauth.ts` owns the OAuth token lifecycle (`resolveOAuthToken`, cache, PKCE flow), while `request.ts` owns helpers that actually send requests. That keeps token resolution reusable without making `oauth.ts` depend back on request execution.

packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -355,6 +367,7 @@ export function createDevServerMiddleware(
     viteBuild: typeof build,
     getBackendFunctions: () => BackendFunction[],
     auth: AuthOptionsWithDefaults,
+    authMethod: AuthMethod,
```
sdkennedy2: Couldn't we just thread through doAuthenticatedRequest like we do in other places?

packages/plugins/apps/src/vite/handle-upload.ts
```diff
@@ -139,12 +151,16 @@ Either:
         archiveDir = path.dirname(archive.archivePath);
 
         const uploadTimer = log.time('upload assets');
+        const doAuthenticatedRequest =
+            (options.dryRun
+                ? doDryRunAuthenticatedRequest
+                : getAuthenticatedRequest(options.authOverrides.method, auth, log)) ||
+            doMissingAuthenticationRequest;
```
sdkennedy2: I wonder if `getAuthenticatedRequest` should just throw if we are missing auth information. We probably need to differentiate between a 500 and 400 error in some cases, perhaps we should create error classes.

packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -130,10 +129,11 @@ async function executeScriptViaDatadog(
     scriptBody: string,
     func: BackendFunction,
     args: unknown[],
-    auth: AuthConfig,
+    site: Site,
```
sdkennedy2: It looks like we missed some places. We were support to revert `site: Site,` changes back to `auth: AuthConfig,`

packages/plugins/apps/README.md
```diff
@@ -31,6 +32,9 @@ apps?: {
     include?: string[];
     identifier?: string;
     name?: string;
+    authOverrides?: {
+        method?: 'apiKey' | 'oauth';
+    };
```
yoannmoinet, line 37: nit: maybe worth mentioning the need to install the optional keyring dependency in the documentation.

packages/plugins/apps/src/vite/index.ts
```diff
@@ -156,10 +165,15 @@ export const getVitePlugin = ({
                 backendFunctions = result.functions;
             }
             try {
+                const doAuthenticatedRequest = options.dryRun
+                    ? doDryRunAuthenticatedRequest
+                    : getAuthenticatedRequest(options.authOverrides.method, auth, log);
```
yoannmoinet, line 170: `MissingAuthenticationError` gets swallowed here.
No logs and no error.
I think adding a log in `getAuthenticatedRequest` would be plenty enough.

packages/core/src/helpers/oauth-request.ts
```diff
@@ -0,0 +1,666 @@
...
+    }
+
+    return undefined;
+};
+
+const createOAuthCredentialEntry = async (
+    site: string,
+    options: Pick<OAuthConfig, 'authorizationUrl' | 'clientId' | 'tokenUrl'>,
+) => {
+    const { AsyncEntry } = await loadKeyring();
```
yoannmoinet, line 407: Should we display a friendly error message hinting at installing the dependency?

packages/core/src/helpers/request.ts
```diff
@@ -93,8 +93,9 @@ export const createRequestData = async (options: {
 
 export const ERROR_CODES_NO_RETRY = [400, 401, 403, 404, 405, 409, 413];
 export const NB_RETRIES = 5;
+
 // Do a retriable fetch.
-export const doRequest = <T>(opts: RequestOpts): Promise<T> => {
+export const doRequest = async <T>(opts: RequestOpts): Promise<T> => {
```
yoannmoinet, line 98: I think the `async` isn't necessary anymore.

yoannmoinet (approved):
Mostly LGTM, just a few comments.
Nothing worth blocking.

packages/plugins/apps/README.md
sdkennedy2: Good call 👍

packages/plugins/apps/src/vite/index.ts
sdkennedy2: Agreed. Good call.

PR #399, opened 2026-06-10
Fix Live Debugger local capture TDZ reads

watson:
### What and why?

Fixes a Live Debugger instrumentation bug where return probes could read local variables before their declaration had executed. The previous transform emitted one function-wide local capture helper, so an early return before a later `let`/`const` declaration could trigger a TDZ error when probes were active.

### How?

Track direct local declarations with their declaration end positions and compute return capture arguments per exit point. Local captures are now inlined at `$dd_return` call sites only when the local declaration is available for that return, while parameter capture still uses the shared entry helper.

Updated transform coverage for inline local captures, no-local returns, and the early-return-before-local edge case. Updated the Live Debugger README examples to remove the old local helper shape.

The current version of this branch is published to npm as version `3.1.8-dev.2`.

### DIFF
--- a/packages/plugins/live-debugger/README.md
+++ b/packages/plugins/live-debugger/README.md
@@ -68,7 +68,7 @@ The Live Debugger plugin automatically instruments all JavaScript functions in y
 Each instrumented function gets:
 - A unique, stable function ID (format: `<file-path>;<function-name>`)
 - A `$dd_probes()` call that returns active probes for that function (or `undefined` if none)
-- Deferred variable-capture helpers (`$dd_e<N>` for entry variables, `$dd_l<N>` for exit variables) that are only evaluated when probes are active
+- Deferred parameter-capture helpers (`$dd_e<N>`)
 - Entry point tracking with parameter capture via `$dd_entry()`
 - Return value tracking with local variable capture via `$dd_return()`
 - Exception tracking with variable state at throw time via `$dd_throw()`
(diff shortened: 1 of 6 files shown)


watson:
* **#399** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/399?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/399?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

packages/plugins/live-debugger/README.md
```diff
@@ -91,16 +91,14 @@ function add(a, b) {
     const $dd_p = $dd_probes('src/utils.js;add');
     const $dd_e = () => ({a, b});
     try {
-        const $dd_l = () => ({a, b, sum});
```
watson, line 94: FYI, the old version of the `README.md` was actually incorrect on this line on several points:
1. The name of the constant was actually `$dd_l0`
2. The return value of `$dd_l0` was actually _only_ the local variables, not also the function params: `{sum}`

This is because the `README.md` had not been kept up to date with a few fixes/refactorings that happended after it was created. That is now taken care of.

etayyosefdd (approved):
LGTM 👍

yoannmoinet (approved):

PR #401, opened 2026-06-10
[RUM-16926] add debug_id snippet injection for cross-MFE stack trace un-minification

amortemousque:
### What and why?

Current unminification based on (service, version) breaks with MFEs.

When an error spans multiple micro frontends, its stack trace may contain frames from several bundles (for example, cart.js, header.js, and shell-app.js). Unminification today relies on a single (service, version) pair, typically derived from the originating MFE. The backend applies that pair to every frame in the stack trace. As a result, frames from other MFEs, which were uploaded under their own (service, version) pairs—cannot be matched to the correct source maps and remain minified.

debug_id solves this by attaching a UUID to each JS bundle. The backend can then resolves each frame independently.

### How?
* Add injection plugin capability to pass the chunk source or hash to hook callbacks.
    * Expose a new injection callback: (sourceOrHash) => {}
    * Move Rollup injection from banner/footer hooks to renderChunk so the plugin can access and modify the generated chunk source.
* Implement debug_id injection in the RUM sourceCodeContext plugin.
    * Subscribe to the new injection callback and generate a deterministic UUID from the file content
    * Add  ddDebugId: uuid to DD_SOURCE_CODE_CONTEXT entries.
* Add end-to-end tests


packages/plugins/injection/src/helpers.ts
amortemousque, line 1: While implementing this, I took the opportunity to simplify the injection logic by:

- `processInjections` + `processItem` becomes `prepareInjections` + `resolveWithFallback` where `prepareInjections` drives the loop  and `resolveWithFallback` resolves async and file injections
- `processInjections` iterated items one at a time with no-await-in-loop. Now `prepareInjections` resolves all injections in parallel.
- Remove `Map<string, ToInjectItem>`: generated UUIDs were never used for lookup, replaced with `ToInjectItem[]` directly.

I think a good potential next step would be to replace the parameterized context.inject API with separate hooks per injection type.

packages/plugins/injection/src/helpers.ts
```diff
@@ -178,3 +147,10 @@ export const isFileSupported = (ext: string): boolean => {
 export const warnUnsupportedFile = (log: Logger, ext: string, filename: string): void => {
     log.warn(`"${yellow(ext)}" files are not supported (${yellow(filename)}).`);
 };
+
+const isPerChunk = (
+    item: ToInjectItem,
+): item is ToInjectItem & { value: (chunk: ChunkInfo) => string } =>
+    typeof item.value === 'function' && item.value.length === 1;
```
amortemousque, line 171: I'm not a fan of this logic to detect `(chunk) => ...` per-chunk functions. 
Improving this area would require a larger refactor, which I’d prefer to tackle in a dedicated PR. My plan is to replace the parameterized context.inject API with separate hooks for each injection type, making the implementation simpler and easier to reason about.

packages/plugins/apps/src/identifier.ts
```diff
@@ -76,9 +76,8 @@ export const buildIdentifier = (repository?: string, name?: string): string | un
     }
 
     const plainIdentifier = `${repository}:${name}`;
-    // Use MD5 hash (128 bits, 32 hex characters) for a compact identifier
-    // MD5 is sufficient for non-cryptographic purposes like creating unique identifiers
-    return createHash('md5').update(plainIdentifier).digest('hex');
+    // SHA-256 truncated to 128 bits (32 hex characters) for a compact identifier.
+    return createHash('sha256').update(plainIdentifier).digest('hex').slice(0, 32);
```
yoannmoinet, line 80: Was this change intentional?

packages/plugins/error-tracking/package.json
```diff
@@ -22,6 +22,7 @@
     "dependencies": {
         "@dd/core": "workspace:*",
         "chalk": "2.3.1",
+        "magic-string": "0.30.21",
```
yoannmoinet: Did we miss it prior?

packages/core/src/types.ts
```diff
@@ -139,7 +147,7 @@ export type ToInjectItem = {
     value: InjectedValue;
     fallback?: ToInjectItem;
 } & (
-    | { position?: InjectPosition.BEFORE | InjectPosition.AFTER; injectIntoAllChunks?: boolean }
+    | { position: InjectPosition.BEFORE | InjectPosition.AFTER; allChunks?: boolean }
```
yoannmoinet: This will need a major bump, IIRC we expose this API through the `customPlugins`.
Is it really necessary?

packages/plugins/injection/src/esbuild.ts
```diff
@@ -128,17 +111,9 @@ export const getEsbuildPlugin = (
...
-                    o.entryPoint && entries.some((e) => e.resolved.endsWith(o.entryPoint!));
-
-                // Get the appropriate banner and footer
-                const banner = isEntry ? bannerForEntries : bannerForAllChunks;
-                const footer = isEntry ? footerForEntries : footerForAllChunks;
-
-                // Skip if nothing to inject for this chunk type
-                if (!banner && !footer) {
-                    continue;
-                }
```
yoannmoinet, line 141: This may have a pretty intense impact on performances, as we will read every single bundle now.
I guess it's necessary for the new injection features, but I would double check the impact on performances still.

If only esbuild would offer us a way to hook ourselves right before writing the bundles.

packages/plugins/error-tracking/src/index.ts
```diff
@@ -5,20 +5,21 @@
...
 
 export { CONFIG_KEY, PLUGIN_NAME } from './constants';
 
 export type types = {
-    // Add the types you'd like to expose here.
     ErrorTrackingOptions: ErrorTrackingOptions;
 };
 
-export const getPlugins: GetPlugins = ({ options, context }) => {
+export const getPlugins: GetPlugins = ({ bundler, options, context }) => {
```
yoannmoinet: `bundler` doesn't seem to be used.

packages/plugins/injection/src/esbuild.ts
yoannmoinet: nit: This was fixed in `rollup` with `MagicString`, would it be a good occasion to do this here too?

packages/plugins/injection/src/helpers.ts
```diff
@@ -62,105 +53,78 @@ export const processLocalFile = async (
...
+    toInject: ToInjectItem[],
+    contentsToInject: ContentsToInject,
     cwd: string = process.cwd(),
-): Promise<
-    Map<string, { injectIntoAllChunks: boolean; position: InjectPosition; value: string }>
-> => {
-    const toReturn = new Map();
-
-    // Processing sequentially all the items.
-    for (const [id, item] of toInject.entries()) {
```
yoannmoinet, line 110: This was required to keep the order of injection the same post processing.

Making it fully parallel may introduce a behavior change worth noting.
This could also lead to indeterministic errors if we have cross injection dependencies.

We could keep the processing parallel, but the final injection should keep the same order to prevent any indeterministic behavior.

wdyt?

packages/plugins/injection/src/index.ts
```diff
@@ -51,11 +50,10 @@ export const getInjectionPlugins: GetInternalPlugins = (arg: GetPluginsArg) => {
             transformIndexHtml: {
                 order: 'pre',
                 handler() {
-                    // For Vite, we inject MIDDLE content by adding a script tag
-                    // that references the virtual injected file
```
yoannmoinet, line 55: Why remove the comment?

packages/published/esbuild-plugin/package.json
```diff
@@ -58,6 +58,7 @@
         "glob": "11.1.0",
         "json-stream-stringify": "3.1.6",
         "jszip": "3.10.1",
+        "magic-string": "0.30.21",
```
yoannmoinet, line 61: Now that it's not optional anymore (it seems) we could remove it from the `peerDependencies`.

packages/plugins/error-tracking/src/debugId.ts
```diff
@@ -0,0 +1,39 @@
...
+
+import { createHash, randomUUID } from 'crypto';
+
+export const SUPPORTED_EXTENSIONS = new Set(['.js', '.mjs', '.cjs']);
+export const DD_DEBUG_IDS = 'DD_DEBUG_IDS' as const;
+
+const VARIANT_CHARS = ['8', '9', 'a', 'b'] as const;
+
+export const getDebugIdSnippet = (codeOrHash?: string): string => {
+    const debugId = codeOrHash ? stringToUUID(codeOrHash) : randomUUID();
```
bcaudan: **❓ question:** ‏What is the reasoning for not alway taking a random id?

packages/plugins/error-tracking/src/types.ts
```diff
@@ -19,10 +19,12 @@ export type SourcemapsOptionsWithDefaults = Required<SourcemapsOptions>;
 
 export type ErrorTrackingOptions = {
     enable?: boolean;
+    debugId?: boolean;
```
bcaudan: **❓ question:** ‏MFE was previously configured under `rum. sourceCodeContext` and we are now adding `errorTracking.debugId`.
Could we limit the number of options that the customers have to enable for MFE supports?

packages/plugins/error-tracking/src/debugId.ts
amortemousque: We want deterministic IDs to prevent duplicates on the backend. Un web UI, most files remain unchanged across releases, and duplicates are very costly.

bcaudan: What about documenting the intent and adding a test on it?

packages/plugins/apps/src/identifier.ts
amortemousque: Yes, intentional. MD5 was only used for deterministic IDs, but it is not supported in FIPS environments, which can cause builds to fail (similar to https://github.com/getsentry/sentry-javascript-bundler-plugins/issues/618). SHA-256 avoids that issue.

packages/core/src/types.ts
amortemousque: Hum, indeed. I can rollback this change to avoid breaking changes

packages/plugins/apps/src/identifier.ts
yoannmoinet: Please notify apps folks about this.
EDIT: Uh oh, it seems I AM part of the apps folks... I'll forward this.

packages/plugins/injection/src/esbuild.ts
amortemousque: Yes it's also one of my concern. I'll get some data on the overhead

packages/plugins/injection/src/esbuild.ts
amortemousque: Yes indeed! It's not exactly the same, as be need to rewrite the already generated .map but I handled them.

packages/plugins/injection/src/index.ts
amortemousque: I don't know why but llms tend to remove them. I'll rewrite them

packages/plugins/injection/src/esbuild.ts
amortemousque: I re-added a check so we inject only when contentsToInject contains an entry with injectIntoAllChunks: true.
This limits the overhead to cases where the customer uses a plugin that requires it.

packages/published/esbuild-plugin/package.json
amortemousque: Indeed :)

packages/plugins/injection/src/helpers.ts
amortemousque: FMU, the order is already preserved. `Promise.all` returns results in input order regardless of resolution timing. Same guarantee as the sequential `for...of`. The only change is parallel execution. Am I missing something?

packages/plugins/error-tracking/src/debugId.ts
amortemousque: Done

packages/plugins/error-tracking/src/types.ts
amortemousque: I collocated the options like to  
```ts
type SourceCodeContext = {
    service?: string;
    version?: string;
    ddDebugId?: string;
};
```
`ddDebugId` is new attribute within DD_SOURCE_CODE_CONTEXT entries

packages/plugins/error-tracking/package.json
amortemousque: Hum, weird I removed it :)

packages/plugins/error-tracking/src/types.ts
bcaudan: Do we still need debugId in ErrorTracking options then?

packages/plugins/rum/src/getSourceCodeContextSnippet.ts
```diff
@@ -19,6 +22,25 @@ export const DEFAULT_SOURCE_CODE_CONTEXT_VARIABLE = 'DD_SOURCE_CODE_CONTEXT' as
...
+    codeOrHash?: string,
+): string => {
+    const context: SourceCodeContext = {
+        service: contextOptions.service,
+        version: contextOptions.version,
+    };
+
+    if (contextOptions.debugId) {
+        // Compute deterministic debug IDs whenever possible preventing the backend from storing duplicate source maps for identical build
+        context.ddDebugId = codeOrHash ? stringToUUID(codeOrHash) : randomUUID();
```
bcaudan, line 45: **💬 suggestion:** ‏if `debugId` VS `ddDebugId` is only to differentiate config option from the actual debug id value, maybe we could rename the option to `enableDebugId` to avoid this `dd` prefix on a single value.

amortemousque: We use ddDebugId so it can be reliably extracted at upload time and avoid false positives from generic debugId fields. That said, I’m open to other suggestions 🙂.

packages/plugins/error-tracking/src/types.ts
amortemousque: Nop!

packages/plugins/rum/src/getSourceCodeContextSnippet.ts
bcaudan: From my understanding, here the debug id is added to the `DD_SOURCE_CEODE_CONTEXT`, how would it conflict with other debugId?

amortemousque: We need a regex to extract the Debug ID at upload time, which can be done by the datadog-ci CLI. To retrieve the value, we read the file and extract it using the regex.
like `const DEBUG_ID_RX = /"ddDebugId":"([^"]+)"/;`

amortemousque: Like discussed, I clarified the intent in a comment: https://github.com/DataDog/build-plugins/pull/401/changes/6f8abf6945400dafc5df2fe19e3ab956dc763095

packages/plugins/rum/src/debugId.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export const SUPPORTED_EXTENSIONS = new Set(['.js', '.mjs', '.cjs']);
+
+// The debug ID is embedded in the DD_SOURCE_CODE_CONTEXT context object as a "ddDebugId" field.
+const DEBUG_ID_RX = /"ddDebugId":"([^"]+)"/;
+
+export const getDebugIdFromSource = (source: string): string | undefined => {
+    const match = source.match(DEBUG_ID_RX);
+    return match ? match[1] : undefined;
+};
```
yoannmoinet: This and `SUPPORTED_EXTENSIONS` are never used, are they really necessary?

packages/plugins/injection/src/esbuild.ts
```diff
@@ -153,15 +128,43 @@ export const getEsbuildPlugin = (
...
+                                return;
+                            }
+
+                            const data = await esbuild.transform(sourceOrHash, {
                                 loader: 'default',
                                 banner,
                                 footer,
+                                sourcemap: hasSourcemap ? 'external' : undefined,
+                                sourcefile: fileName,
                             });
```
yoannmoinet: I believe this will make wrong sourcemaps.
Which will only cover the transition between last transform and this transform.

Shouldn't we "recompute" the existing sourcemaps somehow, to keep them correct?

To have esbuild re-use them, you could make the existing sourcemaps inline by replacing the sourcemaps url with their content directly:

```ts
    const cleaned = sourceOrHash.replace(/\n?\/\/# sourceMappingURL=.*$/m, '');
    const input = hasSourcemap
        ? `${cleaned}\n//# sourceMappingURL=data:application/json;base64,${Buffer.from(existingMap).toString('base64')}`
        : cleaned;

    const data = await esbuild.transform(input, {
        loader: 'default',
        banner,
        footer,
        sourcemap: hasSourcemap ? 'external' : undefined,
        sourcefile: fileName,
    });
```

OR, edit the sourcemaps directly by appending as many `;` as there are lines in the banner, so it shifts the mapping accordingly to the addition.

packages/plugins/injection/src/helpers.ts
yoannmoinet: Haha, good point 🤦
Sorry for the noise.

yoannmoinet (commented):
Just a few comments.

packages/plugins/injection/src/esbuild.ts
amortemousque: You’re right! I tried inlining the source map, and it works well. Nice trick! 🙂
I also added an E2E test to verify that source maps are correctly transformed when using prepending injection.

yoannmoinet (approved):
LGTM. Thanks for all the updates. 
Great sourcemaps test too!

BeltranBulbarellaDD (approved):

bcaudan (approved):
LGTM

PR #403, opened 2026-06-10
Make bump workflow safer to rerun

watson:
### What and why?

Make the manual version bump workflow safer to rerun after partial failures without hiding real branch or tag conflicts from concurrent bump attempts.

### How?

Adds workflow concurrency, which tells GitHub Actions to run only one bump workflow at a time for the same base ref. This prevents two manually triggered bump jobs from creating the same version branch or tag at the same time while keeping later runs queued instead of canceling the active one.

The workflow also checks for pre-existing bump branches and tags before creating remote refs, creates the local bump commit before publishing the remote branch, and tracks which refs were created by the current run so they can be cleaned up if a later step fails. The Datadog log submission is now non-blocking so a logging outage does not mark an otherwise successful bump as failed.

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -12,6 +12,10 @@ on:
           - "minor"
           - "major"
 
+concurrency:
+  group: bump-version-${{ github.ref_name }}
+  cancel-in-progress: false
+
 jobs:
   bump:
     permissions:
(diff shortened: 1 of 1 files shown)


watson:
* **#403** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/403?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/403?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

yoannmoinet (approved):
LGTM, thanks!

PR #404, opened 2026-06-10
fix live debugger sequence arrow body capture

watson:
### What and why?

Fix Live Debugger instrumentation for expression-bodied arrow functions whose body is a comma/sequence expression.

Live Debugger rewrites expression-bodied arrows into block bodies so it can report the function's return value before returning it. To do that, it stores the original expression result in a temporary `$dd_rv` variable, calls the return-reporting hook with that value, then returns the same value.

For example, this valid input:

```ts
const fn = () => (sideEffect(), value);
```

was previously instrumented into invalid JavaScript like this:

```ts
const $dd_rv0 = sideEffect(), value;
```

Because the captured expression was not parenthesized, the comma operator was parsed as a variable declarator separator instead of part of the `$dd_rv0` initializer.

### How?

Detect sequence-expression arrow bodies while collecting transform targets and wrap only those captured return-value initializers in parentheses.

The instrumented output now keeps the sequence expression grouped while still storing its final result in `$dd_rv0` for return reporting:

```ts
const $dd_rv0 = (sideEffect(), value);
```

Added a regression test that verifies the emitted output remains syntactically valid.

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -152,6 +152,17 @@ describe('transformCode', () => {
             expect(result.code).toContain('$dd_probes');
             expect(result.code).toContain('($dd_rv0 = x + 1');
         });
+
+        it('should preserve comma operator semantics in arrow expression body', () => {
+            const result = transformCode({
+                ...BASE_OPTIONS,
+                code: 'const fn = () => (sideEffect(), value);',
+            });
+
+            expect(result.instrumentedCount).toBe(1);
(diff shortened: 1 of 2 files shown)


watson:
* **#404** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/404?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/404?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

yoannmoinet (approved):

PR #405, opened 2026-06-10
Fix nested paren arrow bodies in Live Debugger

watson:
### What and why?

Fix Live Debugger instrumentation for arrow expression bodies wrapped in multiple parentheses. Babel exposes the unwrapped expression range for these bodies, so the previous paren-stripping logic could leave wrapper parens around the injected block and generate invalid JavaScript.

Input examples that now transform correctly:

```ts
const getNumber = () => ((1));
const getObject = () => (({a: 1}));
```

Before this fix, `const getNumber = () => ((1));` could be transformed into invalid syntax shaped like this:

```ts
const getNumber = () => ({
const $dd_p0 = $dd_probes('src/utils.ts;getNumber');
try {
  const $dd_rv0 = 1;
  return $dd_rv0;
} catch(e) {
  throw e;
}
});
```

The injected block was left inside a parenthesized expression, so the first `const` caused a syntax error.

After this fix, the same input is still instrumented and produces a valid block-bodied arrow function shaped like this:

```ts
const getNumber = () => {
const $dd_p0 = $dd_probes('src/utils.ts;getNumber');
try {
  const $dd_rv0 = 1;
  if ($dd_p0) $dd_return($dd_p0, $dd_rv0, this);
  return $dd_rv0;
} catch(e) {
  if ($dd_p0) $dd_throw($dd_p0, e, this);
  throw e;
}
};
```

Single-wrapper cases such as `() => (1)` and `() => ({a: 1})` already worked and continue to work.

### How?

Remove every wrapper paren found between Babel's recorded arrow body range and the containing arrow function end before injecting the block body. Added regression tests covering nested parenthesized scalar and object expression bodies, and verified with `yarn test:unit packages/plugins/live-debugger/src/transform/index.test.ts`.

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -163,6 +163,28 @@ describe('transformCode', () => {
             expect(validateSyntax(result.code, '/src/utils.ts')).toBeNull();
             expect(result.code).toContain('const $dd_rv0 = (sideEffect(), value);');
         });
+
+        it('should instrument arrow expression body wrapped in nested parentheses', () => {
+            const result = transformCode({
+                ...BASE_OPTIONS,
+                code: 'const fn = () => ((1));',
+            });
+
+            expect(result.instrumentedCount).toBe(1);
+            const syntaxError = validateSyntax(result.code, '/src/utils.ts');
+            expect(syntaxError).toBeNull();
+        });
+
+        it('should instrument arrow object expression body wrapped in nested parentheses', () => {
+            const result = transformCode({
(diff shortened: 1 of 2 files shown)


watson:
* **#405** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/405?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/405?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

yoannmoinet (approved):

PR #406, opened 2026-06-11
Fix live debugger this capture before super

watson:
### What and why?

Fix Live Debugger instrumentation for arrows that execute before `super()` completes in derived class constructors. The transform previously injected probe calls that accessed `this`, which can crash valid derived-constructor code while `super()` arguments are still being evaluated.

Before instrumentation, this pattern is valid:

```ts
class B extends A {
  constructor(items) {
    super(items.map((x) => x * 2));
  }
}
```

Before this fix, probing the arrow could inject code that accessed constructor `this` before `super()` returned:

```ts
super(items.map((x) => {
  if ($dd_p0) $dd_entry($dd_p0, this, $dd_e0());
  const $dd_rv0 = x * 2;
  if ($dd_p0) $dd_return($dd_p0, $dd_rv0, this, $dd_e0());
  return $dd_rv0;
}));
```

After this fix, affected arrows use a constructor-scoped `$dd_t` alias that is assigned from `super()`:

```ts
constructor(items) {
  let $dd_t;
  ($dd_t = super(items.map((x) => {
    if ($dd_p0) $dd_entry($dd_p0, $dd_t, $dd_e0());
    const $dd_rv0 = x * 2;
    if ($dd_p0) $dd_return($dd_p0, $dd_rv0, $dd_t, $dd_e0());
    return $dd_rv0;
  })));
}
```

### How?

The transform now detects arrows whose lexical `this` comes from a derived constructor, emits one constructor-scoped `$dd_t` alias, and assigns it from wrapped `super()` calls. Affected arrows pass `$dd_t` to `$dd_entry`, `$dd_return`, and `$dd_throw`; arrows outside derived constructors and non-arrow function expressions continue to use their normal `this`.

Added regression coverage for the reported `super(items.map((x) => x * 2))` pattern, a guard for non-arrow function expressions, and an exact-output smoke test for the generated `$dd_t` shape.

Verified with:
- `yarn test:unit packages/plugins/live-debugger/src/transform/index.test.ts`
- `yarn lint packages/plugins/live-debugger/src/transform/index.ts packages/plugins/live-debugger/src/transform/index.test.ts`
- `yarn cli typecheck-workspaces --files packages/plugins/live-debugger/src/transform/index.ts --files packages/plugins/live-debugger/src/transform/index.test.ts`


watson:
* **#406** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/406?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/406?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tylfin:
@Codex review

packages/plugins/live-debugger/src/transform/index.ts
```diff
@@ -551,13 +594,166 @@ function injectInstrumentation(s: MagicStringType, code: string, target: Functio
...
 
+function injectConstructorThisAlias(
+    s: MagicStringType,
+    code: string,
+    target: ConstructorThisAliasTarget,
+): void {
+    s.update(target.bodyStart, target.bodyStart + 1, `${code[target.bodyStart]}let $dd_t;`);
+
+    for (const superCall of target.superCalls) {
+        s.appendLeft(superCall.start, '($dd_t = ');
```
tylfin, line 626: **Edge case: an arrow whose expression body *is* the `super()` call emits invalid JS.**

When the arrow body is exactly the super call (e.g. `const init = () => super(args);`), the arrow's body range equals the super call range, so the expression-body block-wrap (`s.update(bodyStart, bodyStart + 1, ...)` / `s.update(bodyEnd - 1, bodyEnd, ...)`) and this alias wrap (`appendLeft('($dd_t = '` / `')'`) land on the same boundary positions. Replaying the MagicString ops produces:

```js
() => ($dd_t = {
  const $dd_p0 = $dd_probes('id');
  try { ... const $dd_rv0 = super(); ... }
})
```

`($dd_t = { ... })` parses `{` as an object literal, so the output is a `SyntaxError` and the bundle fails for that file.

The common `super(args.map((x) => ...))` shape is fine because the arrow body sits well inside `super(...)`. The trigger is narrow: only expression-body arrows whose body is the super call itself. It is reachable because `collectSuperCallTargetsFromUnknown` descends into arrows to collect their `super()`.

watson: Fixed 👍

tylfin (approved):

PR #407, opened 2026-06-11
Fix sequence return capture in Live Debugger

watson:
### What and why?

Fix Live Debugger instrumentation for `return` statements whose argument is a sequence expression. Before this change, the instrumentation assigned only the first operand into `$dd_rv`, so `return (a = 2), a + 10` returned `2` instead of `12`, and `return log.push(1), log.push(2), log.length` returned `1` instead of `2`.

### How?

Track whether each collected return argument is a Babel `SequenceExpression`. When injecting return instrumentation, wrap only those sequence arguments in an extra pair of parentheses so `$dd_rv` captures the full sequence result before the probe comma expression is appended.

Added focused regression tests that validate syntax, assert the generated grouping, and execute the transformed code in a small VM sandbox to prove the reported runtime results.

Validated with `yarn test:unit packages/plugins/live-debugger/src/transform/index.test.ts` and `yarn workspace @dd/live-debugger-plugin typecheck`.

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -3,6 +3,7 @@
 // Copyright 2019-Present Datadog, Inc.
 
 import { originalPositionFor, TraceMap } from '@jridgewell/trace-mapping';
+import { runInNewContext } from 'vm';
 
 import { transformCode, validateSyntax } from './index';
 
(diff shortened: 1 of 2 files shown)


watson:
* **#407** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/407?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/407?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tylfin (approved):

PR #409, opened 2026-06-11
Fix Live Debugger probe IDs with quoted method names

watson:
### What and why?

Live Debugger instrumentation could generate invalid JavaScript when a function ID contained a single quote, such as an object method named `"it's"`. This fixes the generated `$dd_probes(...)` call so those function IDs are escaped before being embedded in the injected source.

### How?

Added a small escaping helper for single-quoted JavaScript string literals and use it when generating the probe declaration. Updated the quoted method name regression test to assert the escaped function ID and validate the transformed output syntax.

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -702,16 +702,15 @@ describe('transformCode', () => {
             expect(validateSyntax(result.code, '/src/utils.ts')).toBeNull();
         });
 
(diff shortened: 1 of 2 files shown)


watson:
* **#409** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/409?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/409?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

thenukasiva (approved):

PR #416, opened 2026-06-15
feat(apps): add publish option and DD_APPS_PUBLISH env var

setnilson:
## Motivation

High-code apps had no way to upload a new version without immediately publishing it live. Every upload triggered a `release/live` call, making it impossible to stage a version for review before promotion.

## Changes

Adds a `publish` option to the apps plugin that controls whether `release/live` is called after a successful upload:

- `apps.publish: false` in `vite.config.ts` skips the publish step
- `DD_APPS_PUBLISH=false` / `DATADOG_APPS_PUBLISH=false` env var does the same without touching config
- Explicit config takes precedence over the env var
- Defaults to `true` — existing behavior is unchanged

The companion web-ui PR (DataDog/web-ui#319459) adds an `upload-no-publish` script to the scaffolder so users don't need to know the env var:

```json
"upload": "DD_APPS_UPLOAD_ASSETS=1 vite build",
"upload-no-publish": "DD_APPS_UPLOAD_ASSETS=1 DD_APPS_PUBLISH=false vite build"
```

Files changed:
- `packages/core/src/helpers/env.ts` — registers `APPS_PUBLISH` env var
- `packages/plugins/apps/src/types.ts` — adds `publish` to `AppsOptions` and `AppsOptionsWithDefaults`
- `packages/plugins/apps/src/validate.ts` — resolves `publish` from config or env var
- `packages/plugins/apps/src/upload.ts` — gates the `release/live` PUT call on `context.publish`
- `packages/plugins/apps/src/vite/handle-upload.ts` — passes `options.publish` through
- `packages/plugins/apps/README.md` — documents `apps.publish` and the recommended script pattern
- Tests updated throughout (250 passing)

## QA Instructions

A test app is in `DataDog/experimental` using the published `@datadog/vite-plugin`. Each build bakes in a timestamp and unique badge color so you can tell at a glance whether the live version changed.

```bash
git clone git@github.com:DataDog/experimental.git
cd experimental/user/setnilson/deploy-test-app
npm install
```

**Test 1 — upload and publish in one shot (existing behavior, should still work):**
```bash
dd-auth --actions-api --domain app.datadoghq.com -- npm run upload
```
Open the app in App Builder. Badge color and timestamp should change on reload.

**Test 2 — upload a draft without publishing:**
```bash
dd-auth --actions-api --domain app.datadoghq.com -- npm run upload-no-publish
```
The live badge should **not** change after reload — the draft is saved but not live.

**Test 3 — publish the draft:**

Go to App Builder, open the app, and publish the draft version from the version history. Badge should now update.

## Blast Radius

Only affects apps using `@datadog/vite-plugin` with `apps.enable: true`. Defaults to `true` so no existing upload flow changes. The new env var is opt-in.


packages/plugins/apps/src/upload.ts
```diff
@@ -127,7 +127,9 @@ Would have uploaded ${summary}`,
             log.info(`Your application is available at:\n  ${cyan(appBuilderUrl)}`);
         }
 
-        if (response.version_id) {
+        const shouldPublish = getDDEnvValue('APPS_PUBLISH') !== 'false';
```
sdkennedy2: I notice our existing `APPS_UPLOAD_ASSETS` docs mention we set `APPS_UPLOAD_ASSETS=1` versus here we are using a boolean value. Perhaps we create a helper function for normalizing this so that a customer could set 0/1 or false/true. Then we use the helper when evaluating both env variables.

sdkennedy2 (approved):

PR #421, opened 2026-06-16
Support optional dependency integrity

sdkennedy2:
## Motivation

Published plugin packages currently sync required dependencies from internal workspaces, but the integrity command does not preserve whether a propagated dependency is optional. That makes it difficult to keep native, platform-specific packages optional when they are introduced by shared internal code.

## Changes

- Teach dependency integrity to read internal workspace `optionalDependencies`.
- Propagate expected optional dependencies separately from required dependencies.
- Keep required dependencies authoritative when the same package appears as required in one workspace and optional in another.
- Reuse the same mismatch reporting/apply flow for both dependency sections.

## QA Instructions

```bash
yarn cli integrity
git diff --check
```

### DIFF
--- a/packages/tools/src/commands/integrity/dependencies.ts
+++ b/packages/tools/src/commands/integrity/dependencies.ts
@@ -213,6 +213,18 @@ const syncDependencyRecord = (
     return { dependenciesMatch, newDependenciesToApply, outputLog };
 };
 
+const applyOptionalDependencies = (
+    pkg: PackageJson,
+    optionalDependencies: Record<string, string>,
+) => {
+    if (Object.keys(optionalDependencies).length === 0) {
+        delete pkg.optionalDependencies;
+        return;
+    }
+
+    pkg.optionalDependencies = optionalDependencies;
+};
+
 // Based on the internal dependencies, we need to verify that the declared dependencies are correct
 // in the published packages.
 export const updateDependencies = async (workspaces: Workspace[], bundlers: Workspace[]) => {
@@ -247,6 +259,8 @@ export const updateDependencies = async (workspaces: Workspace[], bundlers: Work
             expected.optionalDependencies,
         );
 
+        let shouldWritePackageJson = false;
+
         if (!dependenciesSync.dependenciesMatch) {
             // Log the error.
             console.log(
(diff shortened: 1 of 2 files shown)


packages/tools/src/commands/integrity/dependencies.ts
```diff
@@ -111,76 +221,50 @@ export const updateDependencies = async (workspaces: Workspace[], bundlers: Work
...
-                outputLog += dim(`\n    "${depName}": "${depVersion}"`);
-            }
+        if (!dependenciesSync.dependenciesMatch) {
+            // Log the error.
+            console.log(
+                `    Mismatch ${red('dependencies')} for ${red(bundler.name)}:\n${dependenciesSync.outputLog}`,
+            );
+            // Fix the dependencies.
+            pkg.dependencies = dependenciesSync.newDependenciesToApply;
+            console.log(`    Writing ${red('package.json')} of ${red(bundler.name)}.`);
```
yoannmoinet: super-nit: this may be printed twice if we're missing both dependencies and optionalDependencies.

packages/tools/src/commands/integrity/dependencies.ts
```diff
@@ -111,76 +221,50 @@ export const updateDependencies = async (workspaces: Workspace[], bundlers: Work
...
-        if (!dependenciesMatch) {
+        if (!optionalDependenciesSync.dependenciesMatch) {
             // Log the error.
             console.log(
-                `    Mismatch ${red('dependencies')} for ${red(bundler.name)}:\n${outputLog}`,
+                `    Mismatch ${red('optionalDependencies')} for ${red(bundler.name)}:\n${optionalDependenciesSync.outputLog}`,
             );
             // Fix the dependencies.
-            pkg.dependencies = newDependenciesToApply;
+            pkg.optionalDependencies = optionalDependenciesSync.newDependenciesToApply;
```
yoannmoinet: nit: could it be that we would rewrite `optionalDependencies: {}` in case it's getting emptied?

yoannmoinet (approved):
LGTM, just two very small nits.

yoannmoinet:
Maybe we could have tests, to fix the implementation too.
Now that AI makes it easy to generate, maybe it's worth doing.
Can also be done in a separate PR too.

sdkennedy2:
Good call on the test. I'll add it 👍

sdkennedy2:
* **#421** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/421?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/421?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

PR #427, opened 2026-06-18
[APPS] Default apps plugin auth to OAuth without keys

sdkennedy2:
## Motivation

Datadog Apps local development and upload should use OAuth by default when API/App keys are not configured. This matches the intended customer setup where `npm run dev` and `npm run upload` can start an OAuth browser flow instead of failing with missing key guidance.

## Changes

The apps plugin now resolves `apps.authOverrides.method` in this order:

```ts
DD_APPS_AUTH_METHOD / DATADOG_APPS_AUTH_METHOD
apps.authOverrides.method
apiKey when both API and App keys are available
oauth otherwise
```

Explicit config and environment overrides still win, and users can force API/App-key auth with `DD_APPS_AUTH_METHOD=apiKey`. The README documents the conditional default, and the validation tests cover configured keys, env keys, incomplete auth, explicit override precedence, and env isolation.

## QA Instructions

In a Datadog App using `@datadog/vite-plugin`, run `npm run dev` or `npm run upload` without `DD_API_KEY`/`DD_APP_KEY`; the apps plugin should select OAuth. Set both keys and rerun to confirm API/App-key auth remains the default. Set `DD_APPS_AUTH_METHOD=apiKey` or `DD_APPS_AUTH_METHOD=oauth` to confirm explicit method selection still wins.

## Blast Radius

This affects the Datadog Apps plugin auth default for app upload and local backend function execution. Other build plugin products keep their existing auth behavior because the override remains scoped to `apps.authOverrides.method`.

## Documentation

### DIFF
--- a/packages/plugins/apps/README.md
+++ b/packages/plugins/apps/README.md
@@ -72,13 +72,15 @@ Additional glob patterns (relative to the project root) to include in the upload
 
 ### apps.authOverrides.method
 
-> default: `apiKey`
+> default: `apiKey` when both `DD_API_KEY` and `DD_APP_KEY` are configured, otherwise `oauth`
 
 Authentication method for uploading app bundles.
 
 Use `apiKey` to send `DD_API_KEY`/`DD_APP_KEY` credentials from the shared `auth` config. Use `oauth` to complete a local Authorization Code + PKCE flow and upload with a short-lived bearer token instead.
 
-You can also set `DATADOG_APPS_AUTH_METHOD=oauth` or `DD_APPS_AUTH_METHOD=oauth`.
+When `apps.authOverrides.method` is not set, the plugin uses API/App-key auth if both keys are configured. If either key is missing, it uses OAuth by default.
+
+You can also set `DATADOG_APPS_AUTH_METHOD` or `DD_APPS_AUTH_METHOD` to `apiKey` or `oauth`.
 
 When the method is `oauth`, the plugin derives OAuth client settings from the resolved Datadog site. The plugin reads tokens from the OS credential store, refreshes expired access tokens when a refresh token is available, and only starts browser authorization when no usable stored token exists.
 
--- a/packages/plugins/apps/src/validate.test.ts
+++ b/packages/plugins/apps/src/validate.test.ts
@@ -3,40 +3,17 @@
 // Copyright 2019-Present Datadog, Inc.
 
 import { validateOptions } from '@dd/apps-plugin/validate';
-
-const authEnvVars = [
-    'DATADOG_API_KEY',
-    'DD_API_KEY',
-    'DATADOG_APP_KEY',
-    'DD_APP_KEY',
-    'DATADOG_APPS_AUTH_METHOD',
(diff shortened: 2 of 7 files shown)


sdkennedy2:
* **#427** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/427?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

sdkennedy2:
@codex review
@cursor review

packages/plugins/apps/README.md
```diff
@@ -72,13 +72,15 @@ Additional glob patterns (relative to the project root) to include in the upload
 
 ### apps.authOverrides.method
 
-> default: `apiKey`
+> default: `apiKey` when both `DD_API_KEY` and `DD_APP_KEY` are configured, otherwise `oauth`
```
otorrillas, line 75: if OAuth is loaded when nothing is configured, then I think the default is `oauth` now.

```suggestion
> default: `oauth`
```

sdkennedy2: Wouldn't it be when `DD_API_KEY` and `DD_APP_KEY` are set, the default is `apiKey`, otherwise the default is `oauth`?

packages/plugins/apps/src/validate.test.ts
```diff
@@ -4,13 +4,47 @@
...
+const restoreAuthEnv = () => {
+    for (const key of authEnvVars) {
+        const value = savedAuthEnv[key];
+        if (value === undefined) {
+            delete process.env[key];
+        } else {
+            process.env[key] = value;
+        }
+    }
+};
```
yoannmoinet: We have this `cleanEnv()` system here: https://github.com/DataDog/build-plugins/blob/ad82fc266c4cdaf0e442eb2dd3813eec0e9ebadb/packages/tests/src/_jest/helpers/env.ts#L68-L99

Currently, we only call it in [beforeAll/afterAll](https://github.com/DataDog/build-plugins/blob/ad82fc266c4cdaf0e442eb2dd3813eec0e9ebadb/packages/tests/src/_jest/setupAfterEnv.ts#L49) but maybe we could extend/modify it to suit your use-case?
Not sure if it would have a real impact if we were to do it on beforeEach/afterEach instead.

(And also make it leverage [OVERRIDE_VARIABLES](https://github.com/DataDog/build-plugins/blob/ad82fc266c4cdaf0e442eb2dd3813eec0e9ebadb/packages/core/src/helpers/env.ts#L29-L38) instead of duplicating them)

yoannmoinet (approved):
LGTM.
I have a comment, but it's not blocking.

Could be a follow-up to avoid duplications and enforce existing patterns.

packages/plugins/apps/src/validate.test.ts
sdkennedy2: That sounds good. I can take a look. Thanks!

PR #428, opened 2026-06-20
docs(live-debugger): add transformation examples and fix README outputs

watson:
### What and why?

The `@dd/live-debugger-plugin` README documented the build-time instrumentation output, but the "After" code samples didn't match what the plugin actually emits, and many transformation edge cases weren't documented at all.

Two concrete inaccuracies in the README examples:
- Local helper variables were shown un-indexed (`$dd_p`, `$dd_e`, `$dd_rv`) when the transform actually emits per-function indexed names (`$dd_p0`, `$dd_e0`, `$dd_rv0`).
- Nit: Captured locals were shown as `{ sum }` (with spaces) when the transform emits `{sum}`.

This makes the docs trustworthy (so readers can correlate generated output with what they see in their bundles) and gives a single place that catalogs the full range of transformation behaviors.

### How?

- Fixed the existing README examples to use indexed helper names and the correct `{sum}` formatting.
- Reduced the README "How it works" section to a single most-common example, followed by a link to the new catalog, to keep the README focused.
- Added `packages/plugins/live-debugger/EXAMPLES.md`: a before/after catalog covering block bodies, arrow expression bodies, object-literal arrow bodies, no-parameter functions, the no-params-but-locals `undefined` placeholder, void functions, bare returns, multiple returns, exhaustive if/else (no trailing return), scope-aware local capture, derived-constructor `this` aliasing (`$dd_t`), arrows in/around `super()`, function expressions keeping their own `this`, the anonymous function ID format, and what is not instrumented.
- Every example is transcribed from the transform's snapshot tests in `src/transform/index.test.ts`; the only computed value (the anonymous-function column offset) was verified directly with `@babel/parser`.
- Ran `yarn cli integrity` to generate the new file's table of contents and validate all cross-file markdown links and anchors.

### DIFF
--- a/packages/plugins/live-debugger/EXAMPLES.md
+++ b/packages/plugins/live-debugger/EXAMPLES.md
@@ -1,0 +1,461 @@
+# Live Debugger Transformation Examples <!-- #omit in toc -->
+
+This page catalogs the before/after output of the Live Debugger transform for a range of
+function shapes and edge cases. For an overview of how the instrumentation works, see the
+plugin's [How it works](./README.md#how-it-works) section.
+
+A few conventions used throughout:
+
+-   The injected globals (`$dd_probes`, `$dd_entry`, `$dd_return`, `$dd_throw`) are provided by the Browser Debugger SDK at runtime — see [Runtime requirements](./README.md#runtime-requirements).
+-   The per-function local helpers carry a numeric suffix (`$dd_p0`, `$dd_e0`, `$dd_rv0`); the index increments for each instrumented function in a file.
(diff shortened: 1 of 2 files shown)


watson:
* **#428** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/428?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/428?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

yoannmoinet (approved):

PR #429, opened 2026-06-22
Add Live Debugger runtime benchmark coverage

watson:
### What and why?

This adds repeatable browser runtime benchmark coverage for dormant Live Debugger instrumentation. The goal is to give Live Debugger browser changes a controlled way to track whether instrumentation changes introduce measurable overhead in real browsers.

### How?

Adds a Playwright benchmark under `packages/tests/src/bench/liveDebuggerRuntime` with Tiny and Hot workload shapes, baseline/control/instrumented variants, and SDK-like dormant probe hooks installed in the page.

Adds a custom reporter that computes conservative per-call overhead bounds, 95% confidence intervals, A/A diagnostics, moving-block bootstrap intervals, autocorrelation diagnostics, browser failures, raw JSON output, and a PR comment body.

Adds `yarn workspace @dd/tests bench:live-debugger:runtime`, a dedicated Playwright config, result artifacts, a non-blocking CI job, a shared Playwright setup action, and Live Debugger contributor docs explaining how to run and interpret the benchmark.

For more details, see the "Runtime benchmark" section in the added [`CONTRIBUTING.md`](https://github.com/DataDog/build-plugins/blob/watson/DEBUG-5787/add-bench/packages/plugins/live-debugger/CONTRIBUTING.md#runtime-benchmark) file in this PR.

### Validation

The benchmark runs in CI as a non-blocking job and uploads raw samples plus the generated report artifacts for review.

### Benchmark note

The current benchmark numbers show a significant dormant-instrumentation overhead in Safari. Most of this is fixed in stacked PR #438.

### DIFF
--- a/.github/actions/setup-playwright-build/action.yaml
+++ b/.github/actions/setup-playwright-build/action.yaml
@@ -1,0 +1,50 @@
+name: Setup Playwright build
+description: Install dependencies, Playwright browsers, and prebuilt plugins for Playwright jobs.
+runs:
+  using: composite
+  steps:
+    - name: Install Node
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version-file: "package.json"
+
+    - name: Cache build:all
+      id: cache-build
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: packages/published/**/dist
+        key: node18-cache-build-${{ hashFiles('packages/core/**', 'packages/factory/**',
+          'packages/plugins/**', 'packages/published/**',
+          'packages/tools/src/**', 'yarn.lock') }}
+
+    - name: Cache playwright binaries
+      id: cache-playwright-binaries
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: |
+          ~/.cache/ms-playwright
+          ~/Library/Caches/ms-playwright
+          %USERPROFILE%\AppData\Local\ms-playwright
+        key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
+
+    - run: yarn install
+      shell: bash
+
+    - name: Install playwright
+      run: yarn workspace @dd/tests playwright install --with-deps
+      shell: bash
+
+    - name: Build all plugins
+      if: steps.cache-build.outputs.cache-hit != 'true'
+      run: yarn build:all-no-types
+      shell: bash
+
+    - name: Save playwright cache
+      if: always() && steps.cache-playwright-binaries.outputs.cache-hit != 'true'
+      uses: actions/cache/save@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: |
+          ~/.cache/ms-playwright
+          ~/Library/Caches/ms-playwright
+          %USERPROFILE%\AppData\Local\ms-playwright
+        key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
--- a/.github/actions/setup-playwright-build/action.yaml
+++ b/.github/actions/setup-playwright-build/action.yaml
@@ -1,0 +1,50 @@
+name: Setup Playwright build
+description: Install dependencies, Playwright browsers, and prebuilt plugins for Playwright jobs.
+runs:
+  using: composite
+  steps:
+    - name: Install Node
+      uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+      with:
+        node-version-file: "package.json"
+
+    - name: Cache build:all
+      id: cache-build
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: packages/published/**/dist
+        key: node18-cache-build-${{ hashFiles('packages/core/**', 'packages/factory/**',
+          'packages/plugins/**', 'packages/published/**',
+          'packages/tools/src/**', 'yarn.lock') }}
+
+    - name: Cache playwright binaries
+      id: cache-playwright-binaries
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: |
+          ~/.cache/ms-playwright
+          ~/Library/Caches/ms-playwright
+          %USERPROFILE%\AppData\Local\ms-playwright
+        key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
+
+    - run: yarn install --immutable --immutable-cache
+      shell: bash
+
+    - name: Install playwright
+      run: yarn workspace @dd/tests playwright install --with-deps
+      shell: bash
+
+    - name: Build all plugins
+      if: steps.cache-build.outputs.cache-hit != 'true'
+      run: yarn build:all-no-types
+      shell: bash
+
+    - name: Save playwright cache
+      if: always() && steps.cache-playwright-binaries.outputs.cache-hit != 'true'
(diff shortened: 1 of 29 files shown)


watson:
* **#429** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/429?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 2  dependent PRs ([#438](https://github.com/DataDog/build-plugins/pull/438) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>, [#440](https://github.com/DataDog/build-plugins/pull/440) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/440?utm_source=stack-comment-icon" target="_blank">Graphite</a>) 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/429?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

tylfin (approved):
LGTM, but I'll defer the CI changes and test configuration to the build-plugins team

.github/workflows/ci.yaml
```diff
@@ -96,37 +74,146 @@ jobs:
...
           path: |
-            ~/.cache/ms-playwright
-            ~/Library/Caches/ms-playwright
-            %USERPROFILE%\AppData\Local\ms-playwright
-          key: cache-playwright-binaries-${{ hashFiles('yarn.lock') }}
+            packages/tests/playwright-report
+            packages/tests/test-results
+          retention-days: 3
+
+  runtime-bench-preflight:
```
yoannmoinet, line 92: Shouldn't this also be `continue-on-error: true` since the other one is?

.github/workflows/ci.yaml
```diff
@@ -96,37 +74,146 @@ jobs:
...
+      - name: Check benchmark and CI changes
+        id: changed-files
+        run: node packages/tests/src/bench/liveDebuggerRuntime/preflight.js --changed-files --base-ref=origin/${{ github.base_ref }}
+
+      - name: Check benchmark output changes
+        if: steps.changed-files.outputs.should-run != 'true'
+        id: output-check
+        run: node packages/tests/src/bench/liveDebuggerRuntime/preflight.js --compare-output --base-ref=origin/${{ github.base_ref }}
+
+  runtime-bench:
```
yoannmoinet, line 128: Since we're posting comments in the PRs, maybe we'll need to avoid runs conflicting (quick succession of pushes).

This way, we only keep the latest one running.


```suggestion
  runtime-bench:
    concurrency: { group: runtime-bench-${{ github.event.pull_request.number }}, cancel-in-progress: true }
```

.github/workflows/ci.yaml
```diff
@@ -96,37 +74,146 @@ jobs:
...
+
+  runtime-bench-preflight:
+    timeout-minutes: 20
+
+    name: Live Debugger runtime benchmark preflight
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request' && !startsWith(github.ref_name, 'mq-working-branch-') }}
+    outputs:
+      should-run: ${{ steps.output-check.outputs.should-run || steps.changed-files.outputs.should-run }}
+      reason: ${{ steps.output-check.outputs.reason || steps.changed-files.outputs.reason }}
```
yoannmoinet: nit: unused.

.github/actions/setup-playwright-build/action.yaml
```diff
@@ -0,0 +1,50 @@
...
+        node-version-file: "package.json"
+
+    - name: Cache build:all
+      id: cache-build
+      uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+      with:
+        path: packages/published/**/dist
+        key: node18-cache-build-${{ hashFiles('packages/core/**', 'packages/factory/**',
+          'packages/plugins/**', 'packages/published/**',
+          'packages/tools/src/**', 'yarn.lock') }}
```
yoannmoinet, line 18: Not directly related to this PR, but it does increase an existing race risk.

Three jobs share the same cache key, `lint`, `e2e` and now `runtime-bench` as well.

Although, `lint` does a build slightly different with `ADD_BUILD_PLUGINS` and `BUILD_PLUGINS_REPORTS`.

Which means, if `e2e` or `runtime-bench` build first, `lint` won't have a report to upload.

Could you change the key for `lint` please?

.github/workflows/ci.yaml
yoannmoinet: nit: Should we also configure Test Optimization for this one too?

packages/tests/playwright.live-debugger-runtime.config.ts
```diff
@@ -0,0 +1,53 @@
...
+ */
+export default defineConfig<TestOptions>({
+    testDir: './src/bench/liveDebuggerRuntime',
+    testMatch: '**/*.bench.ts',
+    fullyParallel: false,
+    workers: 1,
+    forbidOnly: !!process.env.CI,
+    retries: 0,
+    reporter: [['list'], ['./src/bench/liveDebuggerRuntime/reporter/benchReporter.ts']],
+    globalSetup: require.resolve('./src/_playwright/globalSetup.ts'),
```
yoannmoinet: Note that this will conflict with `e2e` runs, because it cleans the public dirs.
So if this runs at the same time as `e2e`, it might clear its fixtures mid-run.

Could be unlikely, but I felt like it was worth a mention.

https://github.com/DataDog/build-plugins/blob/38eda253c9a7539109913137d709d29182ebc5ec/packages/tests/src/_playwright/globalSetup.ts#L37-L42

packages/tests/playwright.live-debugger-runtime.config.ts
```diff
@@ -0,0 +1,53 @@
...
+
+import { PUBLIC_DIR } from '@dd/tests/_playwright/constants';
+import type { TestOptions } from '@dd/tests/_playwright/testParams';
+import { ROOT } from '@dd/tools/constants';
+import { defineConfig, devices } from '@playwright/test';
+
+const BENCH_BUNDLER = 'rspack';
+const BENCH_DEV_SERVER_PORT = 8001;
+const BENCH_DEV_SERVER_URL = `http://localhost:${BENCH_DEV_SERVER_PORT}`;
+const BENCH_BROWSERS = ['chrome', 'firefox', 'safari'] as const;
```
yoannmoinet: These could go in `./constants.ts`.

yoannmoinet (commented):
Some comments.

.github/workflows/ci.yaml
watson: Good catch. I added `continue-on-error: true` to the preflight job so benchmark failures remain non-blocking.

.github/workflows/ci.yaml
watson: Agreed. I added PR-scoped concurrency with `cancel-in-progress` to both benchmark jobs so newer pushes supersede stale runs. I also configured Datadog Test Optimization with the `type:benchmark` tag.

.github/workflows/ci.yaml
watson: Yep, removed 👍

.github/actions/setup-playwright-build/action.yaml
watson: Good catch. Lint now uses a distinct build-report cache key, so it cannot restore a build created without the reporting options.

packages/tests/playwright.live-debugger-runtime.config.ts
watson: I moved these into a new benchmark-local `constants.ts` rather than `@dd/core/constants`, since the browsers, port, URL, bundler, and public directory are specific to this benchmark and aren’t shared core concepts.

packages/tests/playwright.live-debugger-runtime.config.ts
watson: Good point. The benchmark now has its own public directory, and the shared setup only cleans the directory selected by each Playwright configuration.

yoannmoinet (approved):
LGTM, thanks for the updates.

PR #434, opened 2026-06-23
[APPS] Invalidate unusable cached OAuth tokens

sdkennedy2:
## Motivation

Apps OAuth relies on cached tokens from the OS credential store so repeated uploads do not need a browser authorization flow. When that cache contains a bad token, the command can get stuck reusing it instead of recovering.

This PR handles the cases where the cached value is malformed, belongs to a different site/client, has an empty access token, cannot be refreshed, or is rejected by Datadog with an OAuth auth failure.

This is stacked on #435, which defers persistence of newly authorized tokens until the first authenticated request succeeds. This PR adds the follow-up hardening for tokens that already exist in the cache or become unusable later.

## Changes

- Validate cached OAuth credential shape before using it:
  - invalid JSON is deleted from keychain
  - empty access tokens are rejected
  - site/client mismatches are rejected
  - optional token fields must have the expected primitive types
- Delete cached OAuth tokens when refresh fails, so the next command starts browser authorization instead of retrying a stale refresh token.
- Add `RequestError` with HTTP status metadata to the shared request helper.
- When an OAuth-backed request fails with `401` or `403`, clear both the OS credential-store entry and the in-process token memo.
- Keep #435's deferred persistence behavior: refreshed or freshly authorized tokens are persisted only after the OAuth-authenticated request succeeds.
- Update the Apps README to describe that first-time OAuth tokens are saved only after Datadog accepts a plugin request.

## QA Instructions

No manual QA required. The changed behavior is covered by focused unit tests for the OAuth and request helpers.

Validated with:

```sh
yarn workspace @dd/tests test:unit packages/core/src/helpers/oauth-request.test.ts packages/core/src/helpers/request.test.ts --runInBand
```

## Blast Radius

This affects Apps plugin OAuth authentication paths that use the shared core OAuth request helper. API-key and app-key authentication are unchanged.

Expected customer-visible behavior: users with malformed, revoked, wrong-site, wrong-client, expired-unrefreshable, or API-rejected cached OAuth tokens should be prompted to authorize again instead of staying stuck on the unusable cached token.

## Documentation

### DIFF
--- a/packages/core/src/helpers/oauth-request.ts
+++ b/packages/core/src/helpers/oauth-request.ts
@@ -9,7 +9,7 @@ import http from 'http';
 
 import type { AuthOptionsWithDefaults, Logger, RequestOpts } from '../types';
 
-import { doRequest } from './request';
+import { RequestError, doRequest } from './request';
 
 const KEYRING_PACKAGE_NAME = '@napi-rs/keyring';
 
@@ -381,6 +381,10 @@ const getErrorCode = (error: unknown) =>
 const getErrorMessage = (error: unknown) =>
     error instanceof Error ? error.message : String(error);
 
+const isOAuthAuthError = (error: unknown) => {
+    return error instanceof RequestError && (error.statusCode === 401 || error.statusCode === 403);
+};
+
 const isNoEntryError = (error: unknown) => {
     const message = getErrorMessage(error).toLowerCase();
     return (
@@ -391,14 +395,29 @@ const isNoEntryError = (error: unknown) => {
     );
 };
 
-const assertStoredOAuthCredential = (value: unknown): StoredOAuthCredential | undefined => {
+const isOptionalString = (value: unknown) => value === undefined || typeof value === 'string';
+
+const isOptionalNumber = (value: unknown) => value === undefined || typeof value === 'number';
+
+const isNonEmptyString = (value: unknown): value is string =>
+    typeof value === 'string' && value.length > 0;
+
+const assertStoredOAuthCredential = (
+    value: unknown,
+    site: string,
+    options: Pick<OAuthConfig, 'clientId'>,
+): StoredOAuthCredential | undefined => {
     if (
         isObject(value) &&
         value.version === 1 &&
         isObject(value.token) &&
-        typeof value.token.accessToken === 'string' &&
-        typeof value.token.clientId === 'string' &&
-        typeof value.token.site === 'string'
+        isNonEmptyString(value.token.accessToken) &&
+        value.token.clientId === options.clientId &&
+        value.token.site === site &&
+        isOptionalNumber(value.token.expiresAt) &&
+        isOptionalString(value.token.refreshToken) &&
+        isOptionalString(value.token.scope) &&
+        isOptionalString(value.token.tokenType)
     ) {
         return value as StoredOAuthCredential;
     }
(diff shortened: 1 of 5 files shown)


packages/core/src/helpers/oauth-request.ts
```diff
@@ -461,14 +506,14 @@ export const deleteOAuthTokenFromKeychain = async (
     site: string,
     options: Pick<OAuthConfig, 'authorizationUrl' | 'clientId' | 'tokenUrl'>,
 ) => {
+    let entry: Awaited<ReturnType<typeof createOAuthCredentialEntry>>;
```
sdkennedy2, line 509: This refactor lets token reads and explicit token deletes share the same keychain deletion behavior. `deleteOAuthTokenFromKeychain` still ignores missing entries, but the deletion logic now lives in `deleteOAuthCredentialEntry` so `readOAuthTokenFromKeychain` can also clean up malformed or invalid cached credentials without duplicating the `NoEntry` handling. The split also keeps failures to create the keychain entry distinct from failures while deleting the stored password.

packages/core/src/helpers/request.ts
```diff
@@ -147,7 +158,6 @@ export const doRequest = async <T>(opts: RequestOpts): Promise<T> => {
         }
 
         if (!response.ok) {
-            // Not instantiating the error here, as it will make Jest throw in the tests.
```
sdkennedy2, line 150: We shouldn't remove this.

packages/core/src/helpers/oauth-request.ts
```diff
@@ -617,42 +667,36 @@ export const authorizeWithPKCE = async (
     });
 };
 
-export const getOAuthToken = async (
```
sdkennedy2, line 620: This was unused.

packages/core/src/helpers/oauth-request.ts
```diff
@@ -391,14 +395,29 @@ const isNoEntryError = (error: unknown) => {
...
     if (
         isObject(value) &&
         value.version === 1 &&
         isObject(value.token) &&
-        typeof value.token.accessToken === 'string' &&
-        typeof value.token.clientId === 'string' &&
-        typeof value.token.site === 'string'
+        isNonEmptyString(value.token.accessToken) &&
+        value.token.clientId === options.clientId &&
+        value.token.site === site &&
```
sdkennedy2, line 416: If the cached toke is for a different client id or site, we invalidate the cached token.

packages/core/src/helpers/oauth-request.ts
```diff
@@ -671,22 +715,34 @@ export type OAuthRequestOpts = Omit<RequestOpts, 'auth'> & {
...
-    if (persistAfterSuccessfulRequest) {
-        await saveOAuthToken(token, getDatadogOAuthConfig(site), log, site);
-    }
+        if (persistAfterSuccessfulRequest) {
+            await saveOAuthToken(token, options, log, site);
+        }
+
+        return result;
+    } catch (error) {
+        if (isOAuthAuthError(error)) {
```
sdkennedy2, line 739: We now invalidate the stored token on a `401` and `403`

oliverli:
@codex review

packages/core/src/helpers/oauth-request.ts
oliverli: could this be too simple a heuristic? 401/403 could also be caused by granular access issues

packages/core/src/helpers/oauth-request.ts
```diff
@@ -671,22 +715,34 @@ export type OAuthRequestOpts = Omit<RequestOpts, 'auth'> & {
...
+            log.warn(
+                `Datadog OAuth token was rejected by the API; deleting the cached token so the next command starts browser authorization. ${getErrorMessage(error)}`,
```
oliverli, line 741: Tell the user to try again (to trigger a new login with oauth)

oliverli (approved):

PR #438, opened 2026-06-24
Fix Safari dormant overhead in Live Debugger instrumentation

watson:
### What and why?

The Live Debugger runtime benchmark surfaced a large dormant-probe overhead on Safari: a tiny instrumented function cost ~8.9 ns/call (~1,100% over baseline) versus ~0 ns on Chrome and ~1.6 ns on Firefox. The root cause is that the transform hoisted each function's captured parameters into a per-call arrow closure (`const $dd_eN = () => ({...})`). A nested closure that captures parameters forces JavaScriptCore — Safari and all iOS browsers, since they're all WebKit — to heap-allocate the function's scope on every invocation, even when probes are dormant and the closure is never created or called. V8 and SpiderMonkey sink that allocation; JavaScriptCore does not, so it dominated the dormant overhead. Because every instrumented iOS app pays this cost on every call, it's worth eliminating.

### How?

Inline the captured-arguments object literal (`{a, b}`) directly at each probe-guarded call site instead of hoisting it into a per-call closure. The object now lives inside the `if (probe)` / `probe ? ` guards, so dormant calls allocate nothing; gating the closure behind the probe check was confirmed insufficient because the captured scope is still allocated. The active path is unchanged — the same arguments object is built the same number of times when a probe fires. Verified with the runtime benchmark: Safari per-call dormant overhead drops ~73% (8.9 → 2.4 ns) and the Hot workload from ~10 to ~7 ns, with no change on Chrome or Firefox. Transform unit-test snapshots and the plugin README/EXAMPLES output catalog are updated to match the new generated shape.

### DIFF
--- a/packages/plugins/live-debugger/EXAMPLES.md
+++ b/packages/plugins/live-debugger/EXAMPLES.md
@@ -7,8 +7,8 @@ plugin's [How it works](./README.md#how-it-works) section.
 A few conventions used throughout:
 
 -   The injected globals (`$dd_probes`, `$dd_entry`, `$dd_return`, `$dd_throw`) are provided by the Browser Debugger SDK at runtime — see [Runtime requirements](./README.md#runtime-requirements).
--   The per-function local helpers carry a numeric suffix (`$dd_p0`, `$dd_e0`, `$dd_rv0`); the index increments for each instrumented function in a file.
--   The `$dd_e<N>` parameter-capture helper is only emitted when the function has parameters.
+-   The per-function locals carry a numeric suffix (`$dd_p0`, `$dd_rv0`); the index increments for each instrumented function in a file.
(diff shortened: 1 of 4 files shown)


watson:
* **#447** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#442** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#438** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

thenukasiva (approved):

PR #440, opened 2026-06-25
Use the real Browser Debugger SDK in the runtime benchmark

watson:
### What and why?

The Live Debugger runtime benchmark measured dormant overhead against a hand-written stub of the SDK's instrumentation runtime (`$dd_probes`/`$dd_entry`/`$dd_return`/`$dd_throw`), which could silently drift from what actually ships. This PR replaces that stub with the real published Browser Debugger SDK, loaded but kept dormant, so the measured overhead reflects production. It also records per-run provenance — the SDK version and the CDN build fingerprint — so any benchmark result can be tied to the exact build it was measured against.

### How?

- Fetch the published `datadog-debugger.js` bundle from the CDN, inline it before the fixture runs, mock the probe-delivery poll same-origin to return an empty probe set, and call `DD_DEBUGGER.init()` after load so the SDK wires the real `$dd_probes` lookup while staying dormant (no active probes).
- Assert the SDK loaded and is dormant, and read its version from `DD_DEBUGGER.version`, failing the run if it is missing.
- Capture the CDN object's build fingerprint from the bundle-fetch response headers — publish date (`Last-Modified`) and S3 ETag (content hash, normalized from CloudFront's weak `W/"…"` form) — and assert both are present, since the `/v7/` bundle reuses the same release-format version across many master builds and is therefore ambiguous on its own.
- Add an optional `onResponse` hook to `@dd/core`'s `doRequest` so callers can read response headers (it previously returned only the parsed body); the benchmark uses it to read `Last-Modified` and `ETag`.
- Surface the version and fingerprint in the console summary, the PR comment (the hash labeled "S3 ETag" so it is not mistaken for a git commit SHA), and the results JSON, with unit tests covering the label formatting and weak-ETag normalization; update the live-debugger `CONTRIBUTING.md`.

### DIFF
--- a/packages/core/src/helpers/request.ts
+++ b/packages/core/src/helpers/request.ts
@@ -96,7 +96,7 @@ export const NB_RETRIES = 5;
 
 // Do a retriable fetch.
 export const doRequest = async <T>(opts: RequestOpts): Promise<T> => {
-    const { auth, url, method = 'GET', getData, type = 'text' } = opts;
+    const { auth, url, method = 'GET', getData, type = 'text', onResponse } = opts;
     const retryOpts: retry.Options = {
         retries: opts.retries === 0 ? 0 : opts.retries || NB_RETRIES,
         onRetry: opts.onRetry,
@@ -167,6 +167,9 @@ export const doRequest = async <T>(opts: RequestOpts): Promise<T> => {
                 throw new Error(errorMessage);
             }
         }
+
+        // Expose the successful response (headers, status) before the body is consumed.
+        onResponse?.(response);
 
         try {
             let result;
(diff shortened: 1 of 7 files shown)


watson:
* **#440** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#429** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/429?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#438](https://github.com/DataDog/build-plugins/pull/438) <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `master`


<h2></h2>

yoannmoinet (approved):
LGTM

PR #441, opened 2026-06-29
Fix live debugger curried function instrumentation

watson:
### What and why?

Fixes a Live Debugger transform bug where curried functions like this could emit invalid JavaScript:

```ts
const make = (dep) => function named() {
    return dep;
};
```

Before this fix, the outer arrow instrumentation could overwrite the inner function's closing-brace instrumentation. The generated output effectively looked like this, with the inner `try` missing its `catch`/`finally`:

```js
const make = (dep) => {
    const $dd_p0 = $dd_probes('src/utils.ts;make');
    try {
        if ($dd_p0) $dd_entry($dd_p0, this, {dep});
        const $dd_rv0 = function named() {
            const $dd_p1 = $dd_probes('src/utils.ts;named');
            try {
                let $dd_rv1;
                if ($dd_p1) $dd_entry($dd_p1, this);
                return ($dd_rv1 = dep, $dd_p1 ? $dd_return($dd_p1, $dd_rv1, this) : $dd_rv1);
        };
        if ($dd_p0) $dd_return($dd_p0, $dd_rv0, this, {dep});
        return $dd_rv0;
    } catch(e) { if ($dd_p0) $dd_throw($dd_p0, e, this, {dep}); throw e; }
};
```

That invalid output fails downstream parsers with errors like `Missing catch or finally clause` or `Expected a semicolon`.

### How?

The expression-body path now updates the first body character as before, but appends the suffix at `bodyEnd` instead of rewriting `[bodyEnd - 1, bodyEnd]`. This keeps the inner block-body postamble intact when the arrow body is another function or arrow.

Added table-driven nested-function regression coverage for named function expressions, exported function-expression factories, triple-nested callback shapes, and curried arrows across `namedOnly` modes.

PR #442, opened 2026-06-29
Fix Live Debugger returned arrow instrumentation

watson:
### What and why?

Fix a Live Debugger transform bug that could emit invalid JavaScript when instrumenting a block-bodied function that returns an anonymous expression-bodied arrow.

For example, this input:

```js
export const make = (cond) => {
    return () => cond ? a() : b();
};
```

Could generate output shaped like this before the fix, formatted here for readability:

```js
return (
    $dd_rv1 = () => {
        const $dd_p0 = $dd_probes('src/utils.ts;<anonymous>');
        try {
            if ($dd_p0) $dd_entry($dd_p0, this);
            const $dd_rv0 =
                cond ? a() : b(),
                $dd_p1 ? $dd_return($dd_p1, $dd_rv1, this, {cond}) : $dd_rv1
            );

            if ($dd_p0) $dd_return($dd_p0, $dd_rv0, this);
            return $dd_rv0;
        } catch(e) {
            if ($dd_p0) $dd_throw($dd_p0, e, this);
            throw e;
        }
    }
);
```

In all-functions mode, both the inner arrow suffix and the outer return wrapper could be inserted at the same byte offset. The outer return-capture tail landed before the inner arrow suffix closed, leaving invalid JavaScript. Named-only mode avoided the issue by skipping the anonymous inner arrow.

### How?

Change the outer `return EXPR` wrapper tail from `appendLeft` to `appendRight` at `ret.argEnd`, so an inner expression-body arrow suffix already queued at that offset is emitted first and closes the inner instrumentation before the outer return capture tail.

Add regression cases for block-bodied arrows and function declarations returning expression-bodied arrows, including a JSX mapping case, plus a block-bodied inner arrow control. The table continues to parse transformed output and now also asserts return instrumentation is still present.

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -358,6 +358,34 @@ describe('transformCode', () => {
                 expectedInstrumentedCount: 1,
                 expectedTotalFunctions: 2,
             },
+            {
+                description: 'block-body arrow returning an expression-body arrow',
+                code: 'export const make = (cond) => { return () => cond ? a() : b(); };',
+                namedOnly: false,
+                expectedInstrumentedCount: 2,
+                expectedTotalFunctions: 2,
+            },
+            {
+                description: 'function declaration returning an expression-body arrow',
+                code: 'export function build(x) { const y = x + 1; return () => y > 0 ? pos(y) : neg(y); }',
+                namedOnly: false,
+                expectedInstrumentedCount: 2,
+                expectedTotalFunctions: 2,
+            },
+            {
+                description: 'block-body arrow returning an expression-body arrow with JSX mapping',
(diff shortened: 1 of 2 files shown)


watson:
* **#447** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/447?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#442** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#438** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

thenukasiva (approved):

PR #447, opened 2026-07-03
Fix Live Debugger parenthesized arrow body instrumentation

watson:
### What and why?

This fixes two more cases where the Live Debugger transform emitted invalid JavaScript, both in the wrapper-paren removal path for parenthesized arrow expression bodies. They are the same family as the recently fixed nested-function bugs and were found by running the transform over a large battery of adversarial inputs and re-parsing every output.

The first case is a parenthesized arrow body wrapping an instrumented function whose end abuts the closing paren — e.g. `const f = (a) => ((b) => a + b);` — which silently dropped the inner function's entire instrumentation suffix (`return`/`catch`/closing brace) and produced a syntax error. The second is a paren inside a comment within a parenthesized arrow body — e.g. `const f = () => ((x) /* ) */);` — which was mistaken for a wrapper paren and misaligned removal, again producing invalid output.

### How?

The transform strips wrapper parens before converting the arrow to a block body. It did this with MagicString `remove()`, which clears the affected chunk's `intro`/`outro`. When an inner instrumented function had already `appendRight`-ed its suffix to that same offset (because the inner body abuts the closing paren with no whitespace), that suffix was discarded. This swaps `remove()` for `update(pos, pos + 1, '')`, which edits content only and preserves the appended suffix.

It also replaces the two raw paren scanners with a single comment-aware `collectArrowWrapperParens` helper. Only whitespace, comments, and wrapper parens can appear between the outermost paren and the body, so any `/` there must start a comment; skipping `//` and `/* */` spans means parens inside comments are no longer miscounted as wrappers (regex/division cannot appear in that range, so no string/regex handling is needed).

Tests add regression cases for parenthesized inner arrows (nested, double-parenthesized, JSX, and parenthesized-object bodies) and for comment/regex bodies, plus coverage for async functions and class accessors/static members, which previously had no transform-level tests. The full live-debugger unit suite passes (234 tests).

### DIFF
--- a/packages/plugins/live-debugger/src/transform/index.test.ts
+++ b/packages/plugins/live-debugger/src/transform/index.test.ts
@@ -386,6 +386,48 @@ describe('transformCode', () => {
                 expectedInstrumentedCount: 2,
                 expectedTotalFunctions: 2,
             },
+            {
+                description: 'parenthesized inner arrow abutting the closing paren',
+                code: 'const f = (a) => ((b) => a + b);',
+                namedOnly: false,
+                expectedInstrumentedCount: 2,
+                expectedTotalFunctions: 2,
+            },
+            {
+                description: 'parenthesized inner arrow abutting the closing paren (namedOnly)',
+                code: 'const f = (a) => ((b) => a + b);',
+                namedOnly: true,
+                expectedInstrumentedCount: 1,
+                expectedTotalFunctions: 2,
+            },
+            {
+                description: 'double-parenthesized inner arrow',
+                code: 'const f = (a) => (((b) => a + b));',
+                namedOnly: false,
+                expectedInstrumentedCount: 2,
+                expectedTotalFunctions: 2,
+            },
+            {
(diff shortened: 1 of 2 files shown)


watson:
* **#447** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/447?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/447?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#442** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/442?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#438** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/438?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

duviduduvid (approved):

PR #451, opened 2026-07-10
Accept custom subdomains in auth.site, direct OAuth authorize to them

oliverli:
## Summary

- `auth.site` now accepts an optional single-label custom subdomain on top of a known Datadog site (e.g. `myorg.us5.datadoghq.com`), in addition to the existing bare site values (e.g. `us5.datadoghq.com`), which continue to work unchanged.
- Matching is case-insensitive; the resolved site and subdomain are normalized to lowercase.
- The custom subdomain is used to direct the browser to the org's custom URL when starting the OAuth authorization step (`https://<subdomain>.<site>/oauth2/v1/authorize` instead of `https://app.<site>/oauth2/v1/authorize`).
- Any App Builder URL displayed to the user (e.g. "Your application is available at...", "Published uploaded version ... to live.") now reflects the org's actual URL: the backend-provided `app_builder_url` is shown as-is (it's authoritative), and only when the backend doesn't return one do we construct a fallback from the site, the configured subdomain (or `app` if none is set), and `app_builder_id`.
- Token exchange and every other `api.<site>` call (RUM, metrics, error tracking, Apps upload/dev-server) are unaffected — those always use the base site, since custom subdomains are a web-UI alias rather than a machine-API host.

## Details

- Added `parseSite()` in `@dd/core/constants`, which resolves a raw site string into `{ site, subdomain? }`, preferring the most specific (longest) matching base site and rejecting subdomains with multiple labels or invalid characters. Matching is done on a lowercased copy of the input, so `CustomSubdomain.US5.DatadogHQ.com` resolves the same as the all-lowercase form.
- `AuthOptionsWithDefaults` gained an optional `siteSubdomain` field alongside the existing `site` (which stays the strict base-domain union).
- `validateOptions` resolves the `DD_SITE`/`DATADOG_SITE` env var and `options.auth.site` through `parseSite` atomically, so site and subdomain always come from the same source.
- `getDatadogOAuthConfig` takes an optional `subdomain` argument and builds the authorization URL against it when present; this threads through `resolveOAuthToken` and `doOAuthRequest`.
- `uploadArchive` resolves the App Builder URL via `resolveAppBuilderUrl(action, site, subdomain, appBuilderUrl, appBuilderId)`: it returns the backend's `app_builder_url` verbatim when present (never rewritten, even if a custom subdomain is configured), otherwise builds `https://<subdomain ?? 'app'>.<site>/...` from `app_builder_id`. The "could not resolve" warning now only fires when the backend returns neither a URL nor an ID.
- Updated `README.md`, `MIGRATIONS.md`, and `packages/plugins/apps/README.md` to document custom-subdomain support and its OAuth-authorization-specific behavior.

## Test plan

- [x] `yarn test:unit` for `packages/core`, `packages/factory`, `packages/plugins/apps` (all passing), including new cases for bare/subdomain/invalid/mixed-case `auth.site` values, the OAuth authorization URL, and the App Builder URL priority/fallback logic (both upload and release logs).
- [x] `yarn typecheck:all` and `yarn cli integrity` pass cleanly.
- [x] Verified against a locally linked `@datadog/vite-plugin` build in a real Vite project — a full build/apps-upload dry run resolves `auth.site` correctly end to end.
- [x] Not verified against a real Datadog org configured with a custom subdomain — the OAuth library's callback validation compares the authorization server's `issuer` (built from the subdomain host) against the `iss` value Datadog's server returns, which this change doesn't control. Worth a manual check against a real custom-subdomain org before relying on this in production.

### DIFF
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -25,41 +25,6 @@ export const SITES = [
 
 export const DEFAULT_SITE = SITES[0];
 
-// A site value can be a bare known site (e.g. "datadoghq.com") or a single-label
-// custom subdomain on top of one (e.g. "myorg.us5.datadoghq.com"), used by orgs
-// with a custom Datadog URL. Picks the longest (most specific) matching base site
-// so "myorg.us5.datadoghq.com" resolves to "us5.datadoghq.com", not "datadoghq.com".
-// Matching is case-insensitive; the returned site and subdomain are normalized to lowercase.
-export const parseSite = (
-    value: string,
-): { site: (typeof SITES)[number]; subdomain?: string } | undefined => {
-    // Config isn't enforced by the type system at runtime (e.g. plain JS configs),
-    // so a non-string can reach here despite the `string` signature.
-    if (typeof value !== 'string') {
-        return undefined;
-    }
-
-    const lowerValue = value.toLowerCase();
-
-    const exactMatch = SITES.find((s) => s === lowerValue);
-    if (exactMatch) {
-        return { site: exactMatch };
-    }
-
-    const suffixMatches = SITES.filter((s) => lowerValue.endsWith(`.${s}`));
-    if (!suffixMatches.length) {
-        return undefined;
-    }
-
-    const site = suffixMatches.reduce((longest, s) => (s.length > longest.length ? s : longest));
-    const subdomain = lowerValue.slice(0, lowerValue.length - site.length - 1);
-    if (!subdomain || !/^[a-z0-9-]+$/.test(subdomain)) {
-        return undefined;
-    }
-
-    return { site, subdomain };
-};
-
 export const ENV_VAR_REQUESTED_BUNDLERS = 'PLAYWRIGHT_REQUESTED_BUNDLERS';
 
 export const HOST_NAME = 'datadog-build-plugins';
--- a/packages/core/src/constants.ts
+++ b/packages/core/src/constants.ts
@@ -29,22 +29,25 @@ export const DEFAULT_SITE = SITES[0];
 // custom subdomain on top of one (e.g. "myorg.us5.datadoghq.com"), used by orgs
 // with a custom Datadog URL. Picks the longest (most specific) matching base site
 // so "myorg.us5.datadoghq.com" resolves to "us5.datadoghq.com", not "datadoghq.com".
+// Matching is case-insensitive; the returned site and subdomain are normalized to lowercase.
 export const parseSite = (
     value: string,
 ): { site: (typeof SITES)[number]; subdomain?: string } | undefined => {
-    const exactMatch = SITES.find((s) => s === value);
+    const lowerValue = value.toLowerCase();
+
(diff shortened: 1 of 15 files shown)


oliverli:
* **#451** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/451?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/451?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `master`


<h2></h2>

oliverli:
@codex review

oliverli:
@codex review

packages/plugins/apps/src/upload.ts
```diff
@@ -188,17 +212,20 @@ Would have uploaded ${summary}`,
                 },
             });
 
-            if (releaseResponse.app_builder_url) {
+            const releaseAppBuilderUrl = resolveAppBuilderUrl(
+                'release',
+                context.site,
+                context.siteSubdomain,
+                releaseResponse.app_builder_url,
+                releaseResponse.app_builder_id,
```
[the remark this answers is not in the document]
oliverli: Fixed in c4f0dd11 — the release fallback now falls back to the upload response's `app_builder_id` when the release response omits it.

packages/core/src/constants.ts
```diff
@@ -25,6 +25,41 @@ export const SITES = [
...
 export const DEFAULT_SITE = SITES[0];
 
+// A site value can be a bare known site (e.g. "datadoghq.com") or a single-label
+// custom subdomain on top of one (e.g. "myorg.us5.datadoghq.com"), used by orgs
+// with a custom Datadog URL. Picks the longest (most specific) matching base site
+// so "myorg.us5.datadoghq.com" resolves to "us5.datadoghq.com", not "datadoghq.com".
+// Matching is case-insensitive; the returned site and subdomain are normalized to lowercase.
+export const parseSite = (
+    value: string,
+): { site: (typeof SITES)[number]; subdomain?: string } | undefined => {
```
yoannmoinet: I don't think this have its place in a `constants.ts` file.
Also, this kind of helpers, should be well tested.

packages/factory/src/validate.test.ts
```diff
@@ -41,6 +41,61 @@ describe('factory validateOptions', () => {
         });
     });
 
+    describe('auth.site', () => {
```
yoannmoinet, line 47: A lot of these should live alongside `parseSite` only.

oliverli:
@codex review

oliverli:
@codex review

packages/core/src/helpers/site.ts
yoannmoinet, line 1: Looks good, can you add this (and its tests) to the CODEOWNERS as owned by apps too.

yoannmoinet (approved):
The `core` looks good to me.
Only reviewed the `core` code, not the `apps` code.

PR #454, opened 2026-07-13
chore(deps): add missing .yarn/cache entry for patched @jridgewell/remapping

oliverli:
### What and why?

`yarn.lock` already references a patched `@jridgewell/remapping@2.3.5` by checksum (`bd73d9754d...`), but the corresponding `.yarn/cache` zip was never committed — a gap in the repo's Zero-Installs cache. Fresh immutable installs were silently missing this entry and falling back to the network for it.

This was discovered via [a failed bump workflow run](https://github.com/DataDog/build-plugins/actions/runs/29252245412/job/86823425968): the "Push commit" step crashed because the bump locally regenerated this same cache zip and staged it as part of the version-bump commit, and `Asana/push-signed-commits` can't handle binary file diffs (it reads changed files as UTF-8 text). That's kept as expected/desired behavior — the workflow failing when the committed cache is stale is a useful signal, not a bug to suppress — so `bump.yaml` is unchanged here.

### How?

Commit the missing `@jridgewell-remapping-patch-12a506c5d7-bd73d9754d.zip` cache entry so the lockfile and committed cache are back in sync.

PR #455, opened 2026-07-13
ci: enforce yarn cache immutability to catch stale/missing cache entries

oliverli:
### What and why?

CI's `yarn install` steps ran without `--immutable`/`--immutable-cache`, so Yarn would silently fetch any package missing from the committed `.yarn/cache` (falling back to the network) instead of failing. Combined with the "no diff" check in the `lint` job only covering `git diff --exit-code` (tracked-file drift, not new untracked files), a `.yarn/cache` entry could go missing from a commit entirely without any CI job ever noticing.

This is exactly what happened in #454's underlying commit ([238cf08a](https://github.com/DataDog/build-plugins/commit/238cf08a3b8c4b7deeb035d5d0d1cd2d0898f445)): a patched dependency was added to `yarn.lock`, but its `.yarn/cache` zip was never committed, and CI stayed green because installs just quietly re-fetched it from the registry. The same gap also let the `bump.yaml` workflow's own `yarn install` step succeed while silently regenerating a cache zip, which then broke the "Push commit" step with a confusing `UnicodeDecodeError` deep inside `Asana/push-signed-commits` instead of failing clearly, early, and by itself.

Verified locally: removing the missing cache zip and running `yarn install --immutable --immutable-cache` fails with `Cache entry required but missing for @jridgewell/remapping@patch:...`, reproducing the exact gap. With the entry present, the flags have no effect on a normal install.

### How?

Add `--immutable --immutable-cache` to every `yarn install` call in `ci.yaml` and `bump.yaml`:
- `--immutable`: fail if `yarn.lock` would need to change.
- `--immutable-cache`: fail if the `.yarn/cache` folder would need to be created/modified, i.e. if it's missing an entry `yarn.lock` requires.

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -37,7 +37,7 @@ jobs:
           node-version-file: "package.json"
 
       - name: Install dependencies
-        run: yarn install
+        run: yarn install --immutable --immutable-cache
       - name: Bump version
         run: yarn version:all ${{ inputs.version }}
       - name: Add files
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,7 +40,7 @@ jobs:
           api_key: ${{secrets.DATADOG_API_KEY}}
           site: datadoghq.com
 
-      - run: yarn install
+      - run: yarn install --immutable --immutable-cache
 
       - name: Build all plugins
         if: steps.cache-build.outputs.cache-hit != 'true'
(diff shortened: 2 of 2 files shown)


oliverli:
* **#455** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/455?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/455?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#454** <a href="https://app.graphite.com/github/pr/DataDog/build-plugins/454?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `master`


<h2></h2>

yoannmoinet (approved):

PR #465, opened 2026-07-27
fix(apps): parse backend module source instead of reading ModuleInfo#ast

sdkennedy2:
## Problem

Vite 8 swaps its bundler from Rollup to **Rolldown**, which deliberately refuses one thing the Apps plugin depended on:

```
Error: UNSUPPORTED: ModuleInfo#ast
    plugin: 'datadog-apps-plugin'
    hook: 'closeBundle'
```

The backend module-graph collector read `moduleInfo.ast` in `moduleParsed` to derive each backend function's `allowedConnectionIds`. Rolldown keeps that property on its Rollup-compat object but stubs the getter to throw — its AST lives in Rust memory, so materialising it per module isn't something it wants to do. That's a permanent API boundary, not a bug to wait out.

The failure lands badly: the client bundle transforms and writes **successfully**, then the plugin's nested `vite.build()` for the backend function dies in `closeBundle`. Nothing in the error points at your `.backend.ts`, at Vite 8, or at Rolldown.

Reproduced end to end on a real Vite 8.1.5 + Rolldown 1.1.5 build — the versions from the investigation.

## Fix

Parse `moduleInfo.code` with the plugin context's own `parse`. Rolldown withholds only the pre-built AST; the source and resolved dependency IDs are both still there.

```diff
-import type { BaseNode } from 'estree';
@@
+                if (typeof moduleInfo.code !== 'string') {
+                    return;
+                }
+
+                const parsed = this.parse(moduleInfo.code);
                 const record = createParsedModuleRecord(
                     moduleId,
                     buildRoot,
-                    moduleInfo.ast as BaseNode,
+                    parsed,
                     getStaticDependencyIds(moduleInfo).map(normalizeViteModuleId),
                 );
```

Three things keep this small:

- **Using the bundler's own parser is correct by construction.** The bundler already parsed this exact source to compute `importedIds`, so whatever it accepted parses here too. A bundled third-party parser could disagree with the bundler; its own cannot.
- **No TypeScript-capable parser is needed.** `moduleParsed` runs after `transform`, so types and JSX are already gone — verified on the Vite 8 build above, where `'x' as string` arrives as `"x"` and `<div/>` arrives as `_jsx("div", …)`. **Zero new dependencies**; no lockfile or published-package changes.
- **The cast disappears.** `this.parse()` returns a directly assignable node, so the previous `as BaseNode` goes away rather than being replaced.

The null check isn't defensive — `moduleInfo.code` is `string | null`, so it's required to compile. `typeof` rather than `=== null` also covers a bundler leaving the field undefined.

## One extra line worth calling out

`shouldTraverseCollectedModule` now runs *before* the parse. `createParsedModuleRecord` applies the same predicate, but only once an AST exists — and reading a pre-built AST was free, while parsing is not. Without this, the change would parse every `node_modules` module in the backend graph just to discard it. It prevents a performance regression this fix would otherwise introduce.

## Testing

- The existing collector test now drives `code` instead of `ast`.
- **One new regression test**: a `moduleInfo` whose `ast` getter throws `UNSUPPORTED: ModuleInfo#ast`. Verified it fails against `master` with exactly that error, so it genuinely pins the bug.
- Three test helpers updated to supply `code` and a context carrying `parse` — mechanical, but unavoidable.

`yarn test:unit` (1880 passed), `yarn typecheck:all`, `yarn format`, `yarn cli integrity` all clean.

## Deliberately not here

- **Narrowing the Vite peer range to `<= 7.x`** — that was a stopgap for a broken path. With this fixed, Vite 8 works, so narrowing it would regress the support added in `b6ea2c22`. Happy to add a fail-fast capability guard instead if you'd prefer one.
- **Real Rolldown in the test matrix** — the repo pins `vite@6.3.5`; adding Vite 8 touches `yarn.lock` and `runBundlers`, and belongs in its own PR. The regression test pins the failing behaviour meanwhile.
- **Scaffolder version bump** — lives in the `create-apps` repo.

## Separate bug found while investigating

`collectStaticModuleDependencies` mispairs static specifiers with resolved IDs whenever a module imports the same specifier twice, because bundlers report one resolved ID per *unique* specifier. On a 6-specifier module it mispaired three and dropped one, which can attribute a `connectionId` to the wrong file or miss it. **It is unrelated to Vite 8 and present today on Vite 6/7.** Kept out of this PR to stay minimal; the fix and its tests are preserved on the `claude/apps-static-dep-pairing` branch.

### DIFF
--- a/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
+++ b/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
@@ -2,13 +2,13 @@
 // This product includes software developed at Datadog (https://www.datadoghq.com/).
 // Copyright 2019-Present Datadog, Inc.
 
-import type { BaseNode } from 'estree';
 import type { ModuleInfo } from 'rollup';
 import type { Plugin } from 'vite';
 
 import {
     createParsedModuleRecord,
     type ParsedModuleRecord,
+    shouldTraverseCollectedModule,
 } from '../backend/ast-parsing/module-graph';
 
 const VIRTUAL_MODULE_ID_RE = /^(?:\0|virtual:)/;
@@ -30,10 +30,32 @@ export function createBackendModuleGraphCollector(buildRoot: string): BackendMod
                     return;
                 }
 
+                // `createParsedModuleRecord` applies this same predicate, but
+                // only after the AST exists. Checking it here keeps us from
+                // parsing every `node_modules` module just to discard it.
+                if (!shouldTraverseCollectedModule(moduleId, buildRoot)) {
+                    return;
+                }
+
+                // Parse the source instead of reading `moduleInfo.ast`: Rolldown,
+                // the bundler Vite 8 uses by default, stubs that getter to throw
+                // `UNSUPPORTED: ModuleInfo#ast`. `code` is null for external and
+                // synthetic modules.
+                //
+                // `this.parse` is the bundler's own parser, which already parsed
+                // this exact source to compute `importedIds` — so whatever the
+                // bundler accepted parses here too, and no TypeScript-capable
+                // parser is needed (`moduleParsed` runs after `transform`, so
+                // types and JSX are already gone).
+                if (typeof moduleInfo.code !== 'string') {
+                    return;
+                }
+
+                const parsed = this.parse(moduleInfo.code);
                 const record = createParsedModuleRecord(
                     moduleId,
                     buildRoot,
-                    moduleInfo.ast as BaseNode,
+                    parsed,
                     getStaticDependencyIds(moduleInfo).map(normalizeViteModuleId),
                 );
                 if (!record) {
--- a/packages/plugins/apps/src/index.test.ts
+++ b/packages/plugins/apps/src/index.test.ts
@@ -41,17 +41,24 @@ function extractViteTransform(plugins: PluginOptions[]) {
 }
 
 function emitModuleParsed(
-    config: { plugins?: Array<{ moduleParsed?: (moduleInfo: unknown) => void }> },
+    config: {
+        plugins?: Array<{
+            moduleParsed?: (this: { parse: typeof parseAst }, moduleInfo: unknown) => void;
+        }>;
+    },
     id: string,
     code: string,
     importedIds: string[] = [],
 ) {
     for (const plugin of config.plugins ?? []) {
-        plugin.moduleParsed?.({
-            id,
-            ast: parseAst(code),
-            importedIds,
-        });
+        plugin.moduleParsed?.call(
+            { parse: parseAst },
+            {
+                id,
+                code,
+                importedIds,
+            },
+        );
     }
 }
 
--- a/packages/plugins/apps/src/vite/backend-module-graph-collector.test.ts
+++ b/packages/plugins/apps/src/vite/backend-module-graph-collector.test.ts
@@ -6,40 +6,52 @@ import { parseAst } from 'rollup/parseAst';
 
 import { createBackendModuleGraphCollector } from './backend-module-graph-collector';
 
-describe('Backend Functions - backend module graph collector', () => {
-    test('Should collect parsed local module records from Rollup moduleParsed hooks', () => {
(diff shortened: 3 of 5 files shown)


sdkennedy2:
Trimmed after review — the fix is unchanged, the ceremony around it is gone:

- **Deleted `parse-module.ts` and its test** (−85 lines). A separate module plus a dedicated test file to host a two-line function was not worth it; the parse is now two lines in the collector, where the `// why` comment actually belongs.
- **Dropped the duplicated pairing tests** (−71 lines). They re-tested the three scenarios [#467](https://github.com/DataDog/build-plugins/pull/467) already covers at the `createParsedModuleRecord` level, and the collector has no pairing logic of its own — it passes `staticDependencyIds` straight through.
- **Fixed a stale comment** in `type-guards.ts` that still justified itself in terms of TypeScript-ESTree, which no longer appears anywhere in the codebase.

Now **6 files, +267/−53** (was 8 files, +420/−53). Re-verified: 1896 tests pass, the old collector still fails these tests with the exact `UNSUPPORTED: ModuleInfo#ast` error, and `build:all` is clean.

packages/plugins/apps/src/vite/backend-module-graph-collector.ts
```diff
@@ -2,17 +2,27 @@
...
 
 const VIRTUAL_MODULE_ID_RE = /^(?:\0|virtual:)/;
 
+/**
+ * The slice of the plugin context this hook uses. Declared structurally, rather
+ * than using Rollup's `PluginContext`, so the same hook body type-checks against
+ * Rollup's and Rolldown's separately-declared context types.
+ */
+interface ModuleParsedContext {
+    parse: (code: string) => unknown;
```
sdkennedy2: I don't like that this returns `unknown`.

packages/plugins/apps/src/vite/backend-module-graph-collector.ts
```diff
@@ -24,17 +34,49 @@ export function createBackendModuleGraphCollector(buildRoot: string): BackendMod
     return {
         plugin: {
             name: 'dd-backend-module-graph-collector',
-            moduleParsed(moduleInfo: ModuleInfo) {
+            moduleParsed(this: ModuleParsedContext, moduleInfo: ModuleInfo) {
```
sdkennedy2: Should we just pass parse in instead?

sdkennedy2:
Removed leftovers from two approaches this PR went through before landing on the plugin context's `parse`:

- **Reverted `type-guards.ts` entirely.** `ensureProgram`/`isProgramNode` had been widened from `BaseNode` to `unknown` solely to accept `@typescript-eslint/typescript-estree`'s separately-declared node types. That parser is gone, and `this.parse()`'s return is already assignable to `BaseNode` — so the widening had no remaining purpose.
- **Dropped the redundant `ensureProgram` call.** `createParsedModuleRecord` already validates the node on entry (`module-graph.ts:164`); calling it in the collector first was duplicate validation.
- **Dropped the `ModuleParsedContext` interface.** TypeScript infers `this` contextually from Vite's `Plugin` type, which declares `moduleParsed` with `this: PluginContext` — so the explicit annotation was unnecessary. Verified by typechecking without it.
- **Corrected the filter comment.** It claimed filtering before the parse prevents an irrelevant syntax error from failing the build. That is not true: because we now use the bundler's own parser, and the bundler already parsed every module reaching `moduleParsed`, a parse failure here is impossible. The filter is a performance optimisation — it avoids parsing every `node_modules` module a second time — and the comment now says so.

Source diff is down from 63 changed lines across 2 files to **34 across 1**. The functional change is 6 lines; the rest is the filter, the absent-`code` guard, and the rationale comment. 1896 tests pass; typecheck, format and integrity clean.

sdkennedy2:
Superseded by [#468](https://github.com/DataDog/build-plugins/pull/468), which carries the identical change on a properly-named branch (`sdkennedy2/apps-parse-backend-module-source`) with a PR description following the team template. Same commit content: +131/-52 across 5 files.

PR #466, opened 2026-07-28
fix(error-tracking): derive debug_id from file content, not a filename Map

jhssilva:
Follow-up on top of #464.

## Problem

`error-tracking`'s sourcemap upload was silently dropping `debug_id` for
most chunks. Root cause: the RUM plugin computed a debug_id and stored it
in a side-channel `stores.debugIds` Map keyed by the chunk's filename
*at injection time*, before webpack/rspack's `realContentHash` step
renamed the chunk based on its final content. `error-tracking`'s
uploader later looked the value up by the chunk's **final** on-disk
filename, so the pre-rehash key never matched — for a real
module-federation build, this meant ~2659/2660 chunks silently uploaded
with no `debug_id` at all (`JSON.stringify` drops `undefined`).

## Fix

Stop coordinating by filename between the two plugins entirely. The RUM
plugin already embeds the debug_id directly inside each chunk's own JS
content, as a `ddDebugId` literal purpose-built for regex extraction
(see the existing comment in `getSourceCodeContextSnippet.ts`).
`error-tracking`'s uploader now reads that value straight out of the
minified file it's about to upload (`sourcemap.minifiedFilePath`),
instead of trusting any filename as a coordination key. This is immune
to `realContentHash` or any other renaming step a bundler might add.

Changes:
- `packages/plugins/error-tracking/src/sourcemaps/debugId.ts` (new):
  `extractDebugId()`, regex-extracts the debug ID from a file's content.
  Matches both the JSON-quoted (`"ddDebugId":"..."`) and minifier-stripped
  unquoted (`ddDebugId:"..."`) key forms — minifiers like terser drop
  quotes from object keys that are valid identifiers.
- `packages/plugins/error-tracking/src/sourcemaps/sender.ts`: read
  `sourcemap.minifiedFilePath`'s content and extract the debug_id from it,
  instead of a `context.debugIds.get(...)` Map lookup. Removes
  `DebugIdsContext` and the now-unused `toPosixPath` helper.
- `packages/plugins/rum/src/index.ts`: no longer writes to
  `stores.debugIds` — the debug_id already lives in the injected code.
- `packages/core/src/types.ts`, `packages/factory/src/index.ts`,
  `packages/tools/src/helpers.ts`, `packages/tests/src/_jest/helpers/mocks.ts`:
  remove the now-unused `GlobalStores.debugIds` Map and its initializers.

This also fixes the original bug this PR was about (path-separator/basename
key mismatches in `xpack.ts`/`esbuild.ts`), since there's no filename-keyed
Map left to have a key mismatch in.

Verified:
- `packages/plugins/injection`, `packages/plugins/error-tracking`,
  `packages/plugins/rum` jest suites: 454/454 passing, no regressions.
- Real end-to-end build against a module-federation-remote entry
  (`mf_apm_runtime_app`, rspack), uploaded to staging: confirmed each
  nested chunk's uploaded `debug_id` now matches the unique `ddDebugId`
  literal actually embedded in that chunk's own file.

### DIFF
--- a/packages/core/src/helpers/fs.test.ts
+++ b/packages/core/src/helpers/fs.test.ts
@@ -3,9 +3,10 @@
 // Copyright 2019-Present Datadog, Inc.
 
 import { addFixtureFiles } from '@dd/tests/_jest/helpers/mocks';
+import os from 'os';
 import path from 'path';
 
-import { checkFile } from './fs';
+import { checkFile, outputFileSync, readFilePrefix, rmSync } from './fs';
 
 jest.mock('fs/promises', () => {
     const original = jest.requireActual('fs/promises');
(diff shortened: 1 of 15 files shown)


jhssilva:
Superseded by the same branch pushed directly to DataDog/build-plugins (instead of the jhssilva fork), so internal CI (devflow/mergegate / dd-gitlab checks) actually runs on it. See the new PR.

PR #467, opened 2026-07-28
fix(apps): pair static module specifiers with resolved IDs by unique source

sdkennedy2:
## What

`collectStaticModuleDependencies` pairs the resolved dependency IDs the bundler supplies against the static module specifiers found in the AST. It zipped them positionally against *every* specifier — but bundlers report one resolved ID per **unique** specifier string, in first-occurrence order.

So the moment a module imports the same specifier twice, the pairing slips by one: every later specifier is attributed to the wrong dependency, and the last one is dropped. Reproduced against real Rollup output — 6 specifiers, 5 resolved IDs, 3 mispaired:

| specifier | paired to | should be |
| --- | --- | --- |
| `./side` | `side.js` | ✅ |
| `./dup` | `dup.js` | ✅ |
| `./dup` | `reexport.js` | (collapses into the entry above) |
| `./reexport` | `star.js` | `reexport.js` |
| `./star` | `zlast.js` | `star.js` |
| `./zlast` | *dropped* | `zlast.js` |

## Why it matters

These pairings are what map an action-catalog call back to the module it came from, and they feed `importsByVariable`, `exportsByName` and `starExports`. A wrong pairing can attribute a `connectionId` to the wrong file or miss it entirely — and it does so silently, because the build still succeeds. Since `allowedConnectionIds` is an allowlist, a missed ID is a functional break and a wrong one is worse.

This is present today on Vite 6/7 with plain Rollup. It is not related to any bundler compatibility issue.

## Fix

Deduplicate the specifiers by source string, preserving first-occurrence order, so both sequences stay in step.

The deduplication is deliberately by **specifier**, not by resolved ID: `'./dup'` and `'./dup.js'` resolve to the same file but the bundler still reports two entries, and a test pins that they stay separate.

## Testing

- Three pairing cases in `module-graph.test.ts` (duplicate specifiers, distinct specifiers resolving to the same module, and the full side-effect/re-export/star-export ordering case), following that file's existing `parseAst` idiom.
- One test asserting the downstream consequence: `importsByVariable` resolves each binding to the right module.
- Verified all four fail against the current implementation (3 of them; the same-module case passes both ways by design, as the guard against over-deduplicating).
- **New end-to-end test** (`backend-connection-id-collector.test.ts`) driving connection ID collection through a real `vite.build()` and asserting the exact collected IDs — inline literals, transitive constants across modules, and a value behind an `as const` plus a re-export chain. It passes on `master` unchanged, so it is a characterization test: it pins current correct behaviour before anything moves.

That last test closes a real coverage gap. The existing `ast-parsing/` tests feed ASTs in directly, and the `apps_backend_project` integration fixture contains no action-catalog usage at all, so nothing exercised the collector's actual output end to end.

`yarn test:unit`, `yarn typecheck:all`, `yarn format` and `yarn cli integrity` all clean.

## Follow-up

A second PR stacked on this one fixes the Rolldown/Vite 8 incompatibility in the same collector (it reads `moduleInfo.ast`, which Rolldown stubs to throw). Kept separate because this fix stands on its own, changes behaviour for existing users, and shouldn't be blocked on the dependency discussion in that one.

### DIFF
--- a/packages/plugins/apps/src/backend/ast-parsing/module-graph.test.ts
+++ b/packages/plugins/apps/src/backend/ast-parsing/module-graph.test.ts
@@ -270,4 +270,86 @@ describe('Backend Functions - module graph records', () => {
             FOR_OF_CONNECTIONS: { kind: 'unsupported', reason: 'mutated object binding' },
         });
     });
+
+    describe('static dependency pairing', () => {
+        // The caller supplies resolved IDs from the bundler, which reports one
+        // per *unique* specifier string in first-occurrence order. Pairing them
+        // against every specifier in the AST slips by one as soon as a specifier
+        // repeats, which attributes an import to the wrong dependency.
+        const cases = [
+            {
+                description: 'pair duplicate imports of the same specifier once',
+                code: `
+                    import { a } from './dup.js';
+                    import { b } from './dup.js';
+                    import { c } from './other.js';
+                `,
+                staticDependencies: ['/project/src/dup.js', '/project/src/other.js'],
+                expected: [
+                    { source: './dup.js', resolvedId: '/project/src/dup.js' },
+                    { source: './other.js', resolvedId: '/project/src/other.js' },
+                ],
+            },
+            {
+                description: 'keep distinct specifiers that resolve to the same module apart',
+                code: `
+                    import { a } from './dup';
+                    import { b } from './dup.js';
+                `,
+                staticDependencies: ['/project/src/dup.js', '/project/src/dup.js'],
+                expected: [
+                    { source: './dup', resolvedId: '/project/src/dup.js' },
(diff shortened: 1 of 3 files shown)


sdkennedy2:
Closing to keep a single minimal PR against `master` — see [#465](https://github.com/DataDog/build-plugins/pull/465), which is now scoped strictly to the Vite 8 / Rolldown fix (+131/-52 across 5 files).

The specifier-pairing bug this PR fixed is real and unrelated to Vite 8 — it affects Vite 6/7 today. The work is **not lost**: the branch `claude/apps-static-dep-pairing` is preserved with the fix, its unit tests, and the end-to-end `vite.build()` test. Worth reopening or re-cutting separately.

PR #468, opened 2026-07-28
[APPS] fix(apps): build backend functions under Vite 8 / Rolldown

sdkennedy2:
## Motivation

Vite 8 switches its bundler to **Rolldown**, which stubs `ModuleInfo#ast` to throw (its AST lives in Rust memory — a deliberate, permanent API boundary):

```
Error: UNSUPPORTED: ModuleInfo#ast     plugin: 'datadog-apps-plugin'  hook: 'closeBundle'
```

The Apps backend collector read that property to derive each backend function's `allowedConnectionIds`. So **any app with a `.backend.ts` fails to build on Vite 8** — and fails confusingly: the client bundle prints `✓ built`, then the plugin's nested backend build dies in `closeBundle` with nothing pointing at Vite, Rolldown, or your file.

`@datadog/vite-plugin@3.2.8` declares `vite: ">= 5.x <= 8.x"`, so npm installs it on Vite 8 while this documented feature is broken. Trigger is Rolldown, not the Vite major — `rolldown-vite` on Vite 7 hits it too.

## Changes

Parse `moduleInfo.code` with the plugin context's own `parse` instead of reading the bundler's pre-built AST. Rolldown withholds only the AST; the source and resolved dependency IDs are still there.

```diff
-                    moduleInfo.ast as BaseNode,
+                    parsed,
```

One file, ~34 changed lines. Three things worth knowing while reading it:

1. **No new dependencies.** `moduleParsed` runs after `transform`, so types and JSX are already compiled away — no TypeScript-capable parser needed. No lockfile or published-package changes.
2. **The parse is wrapped to fail closed.** `this.parse` is the bundler's parser but *not* its parser configuration — Rollup binds `{ jsx }` to its internal parse but not to `PluginContext.parse`, so it can throw on source the bundler accepted. Latent today, but a module we can't parse could hide a connection ID, so it raises the existing `unsupportedModuleGraphDependency` rather than an opaque `Expression expected`.
3. **`shouldTraverseCollectedModule` moved ahead of the parse.** Reading the old AST was free; parsing isn't. Without this we'd parse every `node_modules` module just to discard it.

The `typeof moduleInfo.code !== 'string'` guard is required to compile (`code` is `string | null`), not defensive padding.

<details>
<summary>Two alternatives implemented and rejected</summary>

- **`@typescript-eslint/typescript-estree`** — eagerly requires `typescript` (an *optional* peer), so it needed a deferred `require` to avoid dragging `typescript` into every consumer of every published plugin. ~50 lines of loader machinery plus a dependency on all five published packages, for capability this path never exercises.
- **Vite's `parseAst` export** — Vite maps its `require` condition to a reduced CJS entry that **omits `parseAst`**, which would break consumers of this plugin's CJS build.

</details>

## QA Instructions

Verified with a real scaffolded app whose backend function calls `@datadog/action-catalog`, with the connection ID in a separate module so resolution must walk the module graph. Every cell measured — pre-fix is published `3.2.8`, post-fix is this PR:

| Path | V7 pre | V7 post | V8 pre | V8 post |
|---|---|---|---|---|
| `vite build` + manifest allowlist | ✅ | ✅ | ❌ | ✅ |
| Dev server → function executes | ✅ | ✅ | ❌ | ✅ |
| Upload → published live | ✅ | ✅ | ❌ | ✅ |

Both ❌ paths fail with `UNSUPPORTED: ModuleInfo#ast`; pre-fix on Vite 8 the upload never reaches the network because `closeBundle` throws first. Vite 7 unchanged throughout. Full logs, versions and method in the [verification comment](https://github.com/DataDog/build-plugins/pull/468#issuecomment-5109354290).

## Blast Radius

Only the Apps backend-function build path. The Apps plugin already refuses non-Vite bundlers, and builds without a `.backend.ts` never reach this code. No dependency changes.

Risk is moderate rather than low for one reason: this collector produces a security-relevant **allowlist**, and silently collecting fewer connection IDs would break backend functions at runtime while the build still succeeded. Hence the fail-closed parse, the untouched specifier pairing, and QA that asserts exact collected IDs rather than exit codes.

## Documentation

---

**Not here, deliberately:** a separate specifier-mispairing bug affecting Vite 6/7 today ([#469](https://github.com/DataDog/build-plugins/pull/469), stacked); narrowing the Vite peer range (unnecessary now that Vite 8 works); adding Rolldown to the automated test matrix (repo pins `vite@6.3.5` — the verification above was manual); the `create-apps` bump (different repo).

### DIFF
--- a/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
+++ b/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
@@ -9,6 +9,7 @@ import {
     createParsedModuleRecord,
     type ParsedModuleRecord,
     shouldTraverseCollectedModule,
+    unsupportedModuleGraphDependency,
 } from '../backend/ast-parsing/module-graph';
 
 const VIRTUAL_MODULE_ID_RE = /^(?:\0|virtual:)/;
@@ -44,12 +45,23 @@ export function createBackendModuleGraphCollector(buildRoot: string): BackendMod
 
                 // Parse the source instead of reading `moduleInfo.ast`: Rolldown,
                 // the bundler Vite 8 uses by default, stubs that getter to throw
-                // `UNSUPPORTED: ModuleInfo#ast`. Using the context's own parser
-                // is also correct by construction — it already accepted this
-                // exact source to compute `importedIds` — and needs no
-                // TypeScript support, since `moduleParsed` runs after
-                // `transform`, so types and JSX are already gone.
-                const parsed = this.parse(moduleInfo.code);
+                // `UNSUPPORTED: ModuleInfo#ast`.
+                //
+                // No TypeScript-capable parser is needed because `moduleParsed`
+                // runs after `transform`, so types and JSX are already compiled
+                // away. Note that `this.parse` is the bundler's parser but not
+                // its parser *configuration* — Rollup binds no options to it,
+                // while its own module parse passes `{ jsx }`. Our nested build
+                // never enables `jsx`, so the two agree today; fail closed rather
+                // than silently if they ever diverge, since a module we cannot
+                // parse may hide a connection ID.
+                let parsed;
+                try {
+                    parsed = this.parse(moduleInfo.code);
+                } catch {
+                    throw unsupportedModuleGraphDependency(moduleId, 'unparseable module source');
+                }
+
                 const record = createParsedModuleRecord(
                     moduleId,
                     buildRoot,
--- a/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
+++ b/packages/plugins/apps/src/vite/backend-module-graph-collector.ts
@@ -37,20 +37,18 @@ export function createBackendModuleGraphCollector(buildRoot: string): BackendMod
                     return;
                 }
 
-                // Parse the source instead of reading `moduleInfo.ast`: Rolldown,
-                // the bundler Vite 8 uses by default, stubs that getter to throw
-                // `UNSUPPORTED: ModuleInfo#ast`. `code` is null for external and
-                // synthetic modules.
-                //
-                // `this.parse` is the bundler's own parser, which already parsed
-                // this exact source to compute `importedIds` — so whatever the
-                // bundler accepted parses here too, and no TypeScript-capable
-                // parser is needed (`moduleParsed` runs after `transform`, so
-                // types and JSX are already gone).
+                // External and synthetic modules have no source to parse.
                 if (typeof moduleInfo.code !== 'string') {
                     return;
                 }
 
+                // Parse the source instead of reading `moduleInfo.ast`: Rolldown,
+                // the bundler Vite 8 uses by default, stubs that getter to throw
+                // `UNSUPPORTED: ModuleInfo#ast`. Using the context's own parser
+                // is also correct by construction — it already accepted this
+                // exact source to compute `importedIds` — and needs no
+                // TypeScript support, since `moduleParsed` runs after
+                // `transform`, so types and JSX are already gone.
                 const parsed = this.parse(moduleInfo.code);
                 const record = createParsedModuleRecord(
                     moduleId,
@@ -58,6 +56,8 @@ export function createBackendModuleGraphCollector(buildRoot: string): BackendMod
                     parsed,
                     getStaticDependencyIds(moduleInfo).map(normalizeViteModuleId),
                 );
+                // Only null when the traversal predicate rejects, which the guard
+                // above already covered; kept to narrow the nullable return type.
                 if (!record) {
                     return;
                 }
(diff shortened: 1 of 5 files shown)


sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review

Note: the previous review covered `ffc0a886`. Commit `ba9de237` has since addressed internal review findings — a non-mutating test helper, removal of an `as` cast in favour of a `typeof` narrowing guard, and a comment split. Please review the head of the branch.

sdkennedy2:
@codex review

Head is now `f65d1ff4`; the last review covered `ba9de237`. Two commits since, both worth a look:
- `c207499d` wraps `this.parse` to fail closed via `unsupportedModuleGraphDependency`, after review disproved the claim that the plugin context's parser matches the bundler's own (Rollup binds `{ jsx }` to its internal parse but not to `PluginContext.parse`).
- `f65d1ff4` makes the filter-before-parse guard observable via a `jest.fn` parser, keeps the parser's own diagnostic in the thrown error, and branches `importedIdResolutions` on `.length` rather than nullishness (Rolldown omits the property entirely).

packages/plugins/apps/src/vite/backend-module-graph-collector.ts
```diff
@@ -54,7 +92,11 @@ function normalizeViteModuleId(id: string): string {
 }
 
 function getStaticDependencyIds(moduleInfo: ModuleInfo): string[] {
-    return moduleInfo.importedIdResolutions?.map(({ id }) => id) ?? [...moduleInfo.importedIds];
+    const resolutions = moduleInfo.importedIdResolutions;
+    if (resolutions?.length) {
+        return resolutions.map(({ id }) => id);
+    }
+    return [...moduleInfo.importedIds];
```
sdkennedy2: Why did this need to change?

sdkennedy2: It didn't — reverted in `ab3b2936`, back to the original `??` form.

I'd added it in response to an internal review note that `??` only fires on null/undefined, so a bundler reporting `importedIdResolutions: []` alongside a populated `importedIds` would silently yield zero static dependencies. I then failed to check whether any bundler actually does that. It doesn't:

- **Rollup** exposes a real getter — `Array.from(sourcesWithAttributes.keys(), source => module.resolvedIds[source]).filter(Boolean)` (`rollup.js:18382`). It is only empty when the module has no static imports, in which case `importedIds` is empty too, so both forms return `[]`.
- **Rolldown** omits the property entirely (it is absent, not `[]`), so both forms fall through to `importedIds`.

So the two versions are indistinguishable across the whole declared peer range. It was speculative hardening on an untouched context line, in a PR scoped to the Rolldown fix — exactly the kind of thing that shouldn't be riding along. Good catch.

sdkennedy2:
## Corrected coverage table — every cell now measured before *and* after the fix

The earlier table only exercised the dev-server and upload paths **with** the fix, so nothing in it actually demonstrated the fix mattered for those paths. Filled in the missing controls. All runs against production `datadoghq.com`, app on Vite 7.3.6 / 8.1.5, same fixture, same backend function.

"pre-fix" = published `@datadog/vite-plugin@3.2.8` from npm. "post-fix" = this PR, packed and installed as a real tarball so it resolves the app's Vite rather than its own.

| Path | Vite 7 pre-fix | Vite 7 post-fix | Vite 8 pre-fix | Vite 8 post-fix |
|---|---|---|---|---|
| `vite build` + manifest allowlist | ✅ | ✅ | ❌ `UNSUPPORTED: ModuleInfo#ast` | ✅ |
| Dev server → function executes | ✅ | ✅ | ❌ `UNSUPPORTED: ModuleInfo#ast` | ✅ |
| Upload → published to live | ✅ `4494a99dbebde29f` | ✅ `0bfcea55c5f1556f` | ❌ fails in build, upload never reached | ✅ `3f973f03f38b5efe` |

Three things this establishes that the previous table did not:

1. **The dev-server path is broken pre-fix under Vite 8 too**, not just `vite build`. That matters because the dev server runs the same collector through a separate nested build (`dev-server.ts` → `createBackendConnectionIdCollector`), so it was a genuinely independent path. A developer on Vite 8 could not run their app locally either.
2. **Pre-fix on Vite 8, the upload never happens at all** — `closeBundle` throws during the backend build, so nothing is archived or transmitted. The failure is total, not partial.
3. **Vite 7 is byte-for-byte unaffected**, verified on all three paths rather than assumed. Under Rollup, `PluginContext.parse` is the same `parseAst` the old `moduleInfo.ast` getter used, so this is expected — but it is now measured.

Each of the four ✅ upload cells is a distinct published version in the app's history, so the runs are independently auditable.

sdkennedy2:
@codex review

otorrillas (approved):

PR #469, opened 2026-07-28
[APPS] fix(apps): wrong connection IDs from mispaired module specifiers

sdkennedy2:
> **Stacked on [#468](https://github.com/DataDog/build-plugins/pull/468)** — review that first; this PR targets its branch, so the diff here is only the pairing fix.

## Motivation

`collectStaticModuleDependencies` pairs the bundler's resolved dependency IDs against the AST's static specifiers **by index**. But bundlers report one entry per *unique* dependency, so a module that imports the same specifier twice yields more specifiers than IDs and the pairing slips by one from there on.

Reproduced against a real Rollup build — 6 specifiers, 5 resolved IDs, 3 mispaired and 1 dropped:

| specifier | paired to | should be |
| --- | --- | --- |
| `./dup` (2nd) | `reexport.js` | *(collapses into the entry above)* |
| `./reexport` | `star.js` | `reexport.js` |
| `./star` | `zlast.js` | `star.js` |
| `./zlast` | *dropped* | `zlast.js` |

**Why it matters:** these pairings map an action-catalog call back to its module and feed `importsByVariable` / `exportsByName` / `starExports`. A wrong pairing can attribute a `connectionId` to the wrong file or miss it. Since `allowedConnectionIds` is an allowlist, a miss breaks the function at runtime — and the build still exits 0, so it fails **silently**.

Live today on Vite 6/7 with plain Rollup. Unrelated to any bundler compatibility issue; it only shares a code path with #468.

## Changes

Deduplicate the specifiers by source string, preserving first-occurrence order, so both sequences stay in step.

```ts
function getStaticModuleSources(ast: Program): string[] {
    const sources = new Set<string>();
    for (const node of ast.body) {
        // ...import / export-from declarations
        sources.add(node.source.value);
    }
    return [...sources];
}
```

**Known gap, documented rather than hidden.** The two bundlers disagree on what "unique" means — verified against real builds, not assumed:

| One file imported via `./dup.js` *and* `./dup` | Entries reported |
| --- | --- |
| Rollup 4 (Vite 7) | **2** — dedups by specifier string |
| Rolldown 1.1.5 (Vite 8) | **1** — dedups by resolved module ID |

They agree unless one module imports the same file through two different specifiers. In that case this fix is correct under Rollup and **still mispairs under Rolldown**. Closing that needs pairing by *path correspondence* instead of position — a larger change, so it's called out in the code comment and in the test that encodes the Rollup-specific expectation.

I also tried failing closed on a specifier/ID count mismatch. It broke 18 existing tests, because legitimate callers pass partial dependency lists — a build error there would be worse than the narrow remaining gap.

## QA Instructions

No manual QA needed. Four unit cases cover the pairing (three fail without the fix), one asserts `importsByVariable` resolves each binding to the right module, and a new end-to-end test drives connection-ID collection through a real `vite.build()` asserting the exact collected IDs — it passes on `master` unchanged, so it characterises existing behaviour before the fix moves it.

## Blast Radius

Only the Apps backend-function build path — how collected module records map imports to resolved modules, which feeds `allowedConnectionIds`.

**This changes behaviour for existing Vite 6/7 users**, which is the point: allowlists that were silently wrong become correct. Any app whose backend graph imports one specifier twice could previously have had a missing or misattributed connection ID. No dependency or public API changes.

Risk: moderate — it touches the allowlist path, so a mistake has runtime consequences. Mitigated by the tests above plus the end-to-end assertion of exact IDs.

## Documentation

### DIFF
--- a/packages/plugins/apps/src/backend/ast-parsing/module-graph.test.ts
+++ b/packages/plugins/apps/src/backend/ast-parsing/module-graph.test.ts
@@ -270,4 +270,91 @@ describe('Backend Functions - module graph records', () => {
             FOR_OF_CONNECTIONS: { kind: 'unsupported', reason: 'mutated object binding' },
         });
     });
+
+    describe('static dependency pairing', () => {
+        // The caller supplies resolved IDs from the bundler, which reports one
+        // per *unique* dependency in first-occurrence order. Pairing them against
+        // every specifier in the AST slips by one as soon as a specifier repeats,
+        // which attributes an import to the wrong dependency.
+        const cases = [
+            {
+                description: 'pair duplicate imports of the same specifier once',
+                code: `
+                    import { a } from './dup.js';
+                    import { b } from './dup.js';
+                    import { c } from './other.js';
+                `,
+                staticDependencies: ['/project/src/dup.js', '/project/src/other.js'],
+                expected: [
+                    { source: './dup.js', resolvedId: '/project/src/dup.js' },
+                    { source: './other.js', resolvedId: '/project/src/other.js' },
+                ],
+            },
+            {
+                // Rollup keys its resolved-ID list by specifier string, so it
+                // reports two entries here. Rolldown keys by resolved module ID
+                // and reports one — verified against real builds of both — so
+                // this shape still mispairs under Rolldown. Fixing that needs
+                // pairing by path correspondence instead of position.
+                description: 'keep distinct specifiers that resolve to the same module apart',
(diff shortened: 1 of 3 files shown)


sdkennedy2:
@codex review
@cursor review

sdkennedy2:
@codex review

Head is now `cb0af18e` (last reviewed `8db13595`); it was restacked after the parent branch changed.

PR #470, opened 2026-07-28
[dop] [error-tracking]: derive debug_id from file content

jhssilva:
Summary
----

The RUM plugin injects a `debug_id` into every built chunk's own JS content. This PR makes the `error-tracking` plugin extract that `debug_id` back out of each file when it uploads that file's sourcemap, instead of relying on the previous filename-based coordination between the two plugins.

Why
----

Previously, `error-tracking` got the `debug_id` from a side-channel Map that the RUM plugin populated at injection time, keyed by the chunk's filename. Bundlers (webpack/rspack) rename chunks after injection via `realContentHash`, so by the time `error-tracking` looked the value up by the chunk's final filename, the key no longer matched — the `debug_id` was silently dropped for almost every chunk in a real build.

How
----

Instead of coordinating through a filename-keyed Map, `error-tracking` now reads each minified file's own content directly and extracts the `debug_id` from the `ddDebugId` literal the RUM plugin already embeds inside it. This works no matter what renaming the bundler does afterward, since the debug_id travels with the file's content, not its name.

- `error-tracking/src/sourcemaps/debugId.ts` (new): `extractDebugId()` — regex-extracts the debug_id from a file's content.
- `error-tracking/src/sourcemaps/sender.ts`: reads the minified file's content and extracts the debug_id from it, in place of the old Map lookup.
- `plugins/rum/src/index.ts` + `getSourceCodeContextSnippet.ts`: RUM no longer needs to track debug_ids separately — the value already lives in the code it injects.
- `core/src/helpers/fs.ts`: new `readFilePrefix()` helper, since we only need the first bytes of each file, not the whole thing.
- `plugins/injection/{xpack,esbuild}.ts`: keep injected chunk filenames as relative paths for consistency with the rest of the pipeline.

Notes
----
- Real end-to-end build against a module-federation-remote entry (`mf_apm_runtime_app`, rspack), uploaded to staging: each nested chunk's uploaded `debug_id` matches the `ddDebugId` literal embedded in that chunk's own file.

### DIFF
--- a/packages/core/src/helpers/fs.ts
+++ b/packages/core/src/helpers/fs.ts
@@ -76,20 +76,6 @@ export const readFile = (filepath: string) => {
     return fsp.readFile(filepath, { encoding: 'utf-8' });
 };
 
-// Read only the first `maxBytes` bytes of a file.
-// Useful when the data we need is guaranteed to live near the start of the file
-// and reading the whole (potentially large) file into memory would be wasteful.
-export const readFilePrefix = async (filepath: string, maxBytes: number): Promise<string> => {
-    const fd = await fsp.open(filepath, 'r');
-    try {
-        const buffer = Buffer.alloc(maxBytes);
-        const { bytesRead } = await fd.read(buffer, 0, maxBytes, 0);
-        return buffer.toString('utf-8', 0, bytesRead);
-    } finally {
-        await fd.close();
-    }
-};
-
 export const readFileSync = (filepath: string) => {
     return fs.readFileSync(filepath, { encoding: 'utf-8' });
 };
--- a/packages/core/src/helpers/fs.ts
+++ b/packages/core/src/helpers/fs.ts
@@ -76,6 +76,20 @@ export const readFile = (filepath: string) => {
     return fsp.readFile(filepath, { encoding: 'utf-8' });
 };
 
+// Read only the first `maxBytes` bytes of a file.
+// Useful when the data we need is guaranteed to live near the start of the file
+// and reading the whole (potentially large) file into memory would be wasteful.
+export const readFilePrefix = async (filepath: string, maxBytes: number): Promise<string> => {
+    const fd = await fsp.open(filepath, 'r');
+    try {
+        const buffer = Buffer.alloc(maxBytes);
+        const { bytesRead } = await fd.read(buffer, 0, maxBytes, 0);
+        return buffer.toString('utf-8', 0, bytesRead);
+    } finally {
+        await fd.close();
+    }
+};
+
 export const readFileSync = (filepath: string) => {
     return fs.readFileSync(filepath, { encoding: 'utf-8' });
 };
--- a/packages/plugins/error-tracking/src/sourcemaps/sender.ts
+++ b/packages/plugins/error-tracking/src/sourcemaps/sender.ts
@@ -3,7 +3,7 @@
 // Copyright 2019-Present Datadog, Inc.
 
 import { getDDEnvValue } from '@dd/core/helpers/env';
-import { getFile, readFilePrefix } from '@dd/core/helpers/fs';
+import { getFile } from '@dd/core/helpers/fs';
 import {
     createRequestData,
     doRequest,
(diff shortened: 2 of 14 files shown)


packages/plugins/injection/src/xpack.ts
```diff
@@ -134,7 +134,12 @@ export const getXpackPlugin =
                             continue;
                         }
 
-                        const fileName = path.basename(file);
+                        // `file` is already relative to the output directory (and may
+                        // include subdirectories for nested chunks), matching the
+                        // relativePath computed for sourcemap uploads in
+                        // error-tracking/src/sourcemaps/files.ts. Do not collapse it to
+                        // a basename or debug-id lookups by relative path will never match.
+                        const fileName = file.split(path.sep).join('/');
```
buranmert: the same concern

packages/plugins/error-tracking/src/sourcemaps/sender.ts
```diff
@@ -209,9 +200,22 @@ export const sendSourcemaps = async (
...
+    const payloadsQueue = new Queue({ concurrency: options.maxConcurrency });
+    const payloads: Payload[] = await payloadsQueue.addAll(
+        sourcemaps.map((sourcemap) => async () => {
+            // Read the debug_id straight from the minified file's own content instead of
+            // trusting a filename as a coordination key with the RUM plugin — the bundler may
+            // still rename the file after injection (e.g. webpack/rspack's realContentHash),
+            // but the content, and the debug_id embedded in it, is unaffected. Only the file's
+            // prefix is read, since the RUM plugin's injected snippet always lands near the
+            // start of the file (see DEBUG_ID_SEARCH_PREFIX_BYTES).
+            const fileContent = await readFilePrefix(
```
rgaignault: Suggestion: With esbuild, this runs before the injection onEnd, so it uploads the pre-injection files without ddDebugId. Should't we defer the upload until injection completes.

packages/core/src/helpers/fs.ts
```diff
@@ -76,6 +76,20 @@ export const readFile = (filepath: string) => {
...
+export const readFilePrefix = async (filepath: string, maxBytes: number): Promise<string> => {
+    const fd = await fsp.open(filepath, 'r');
+    try {
+        const buffer = Buffer.alloc(maxBytes);
+        const { bytesRead } = await fd.read(buffer, 0, maxBytes, 0);
+        return buffer.toString('utf-8', 0, bytesRead);
+    } finally {
+        await fd.close();
+    }
+};
```
bcaudan: **💬 suggestion:** ‏if we don't plan other usages for this helper, what about moving that to debugId and only expose a:

```ts
function extractDebugId(filePath: string) {
...
}
```

packages/plugins/error-tracking/src/sourcemaps/sender.ts
jhssilva: For now, we'll be dogfooding in web-ui. We'll do other iterations as needed in future pr's.
We'll keep this in mind.

packages/plugins/injection/src/esbuild.ts
```diff
@@ -133,7 +133,16 @@ export const getEsbuildPlugin = (
...
+                            // Keep the path relative to the output directory (subdirectories
+                            // included) so it matches the relativePath computed for sourcemap
+                            // uploads in error-tracking/src/sourcemaps/files.ts. A bare
+                            // basename would break debug-id lookups for nested chunks.
+                            const fileName = context.bundler.outDir
+                                ? path
+                                      .relative(context.bundler.outDir, absolutePath)
+                                      .split(path.sep)
+                                      .join('/')
+                                : path.basename(absolutePath);
```
buranmert: why do we do this? i think this change is beyond the scope of this PR, no?

packages/core/src/helpers/fs.ts
buranmert: i agree

sdkennedy2 (approved):

PR #491, opened 2026-08-21
[apps][long-polling] make retries configurable & add jitter, exponential backoff and a per-attempt abort

Ayc0:
## Motivation

The dev server's long-poll loop was hardcoded to 10 attempts that fired back-to-back with no delay, and nothing bounded a single attempt — if a connection stalled it just hung. Made the retry behaviour configurable and added the two standard API auto-retry strategies (jitter + exponential backoff) on top.

## Changes

- New `apps.longPolling` config: `maxRetries` (set to `1` to disable long-poll retrying entirely), `jitter`, `exponentialBackoff`, `timeoutMs`.
- Each attempt now carries an `AbortSignal.timeout(timeoutMs)`. A stalled attempt is abandoned and **re-polled against the same receipt** rather than failing the action — the receipt stays valid across attempts.
  - `timeoutMs` defaults to 40s, not 30s: the deadline needs headroom over the server's ~30s window, otherwise a healthy poll racing its own response gets aborted.
- Retry delay is capped at 2s and uses equal jitter (`d/2 + random*d/2`). `done: false` is the *expected* outcome of a healthy poll, not a failure, so the delay is there to de-synchronize concurrent pollers — every ms of it is time with no poll in flight.
- `RequestOpts` gains a `signal` that `doRequest` forwards to `fetch` (also flows through the OAuth path).

Two things worth flagging for review, both caught after the first pass:

- `doRequest` calls `bail(error)` on any fetch-level rejection, so the naive version of this (just passing a signal) made things *worse* than no signal at all — a stall returned a 500 on attempt 1 instead of retrying. The request is now wrapped so aborts `continue` and non-abort errors rethrow.
- The abort check can't use `instanceof Error`. Node rejects with a `DOMException` built in undici's realm, which fails `instanceof` across realm boundaries (vm contexts, and the Jest env) — so the retry silently wouldn't happen. It matches on `name` structurally instead.

## QA Instructions

No visual change — dev-server behaviour only.

Covered by unit tests, including the paths that were broken: stall-then-retry against a real abort, `maxRetries: 1` fail-fast, and non-abort errors surfacing instead of being retried away.

## Blast Radius

`apps` dev server only (`/__dd/executeAction`), which is alpha. Defaults keep the previous 10-attempt behaviour, so the only change for existing users is that retries are now spaced out and a hung poll recovers instead of hanging.

The `signal` addition to `RequestOpts` is opt-in and unset everywhere else, so no other product plugin changes behaviour.

### DIFF
--- a/packages/plugins/apps/src/vite/dev-server.ts
+++ b/packages/plugins/apps/src/vite/dev-server.ts
@@ -56,7 +56,7 @@ function isAbortError(error: unknown): boolean {
 // Equal jitter (half fixed, half random) so the delay keeps a floor.
 function getRetryDelay(attempt: number, config: LongPollingConfig): number {
     const backoffDelay = config.exponentialBackoff
-        ? Math.min(RETRY_BASE_DELAY_MS * 2 ** attempt, RETRY_MAX_DELAY_MS)
+        ? Math.min(RETRY_BASE_DELAY_MS * 2 ** (attempt - 1), RETRY_MAX_DELAY_MS)
         : RETRY_BASE_DELAY_MS;
 
     return config.jitter ? backoffDelay / 2 + Math.random() * (backoffDelay / 2) : backoffDelay;
--- a/packages/plugins/apps/src/vite/dev-server.ts
+++ b/packages/plugins/apps/src/vite/dev-server.ts
@@ -32,10 +32,8 @@ const DEV_VIRTUAL_PREFIX = 'virtual:dd-backend-dev:';
 type AuthConfig = AuthOptionsWithDefaults;
 type LongPollingConfig = Required<LongPollingOptions>;
 
-// Kept small on purpose: a `done: false` response is the expected outcome of a
-// healthy poll, not a failure, and any delay here is time with no poll in
-// flight. The delay exists to de-synchronize concurrent pollers, not to back
-// off a broken endpoint.
+// Kept small: `done: false` is healthy, so this delay is dead time. It only
+// exists to de-synchronize concurrent pollers.
 const RETRY_BASE_DELAY_MS = 250;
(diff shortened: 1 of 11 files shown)


packages/plugins/apps/src/vite/dev-server.ts
```diff
@@ -29,6 +30,37 @@ type BundleFn = (func: BackendFunction) => Promise<BundleResult>;
...
+        return false;
+    }
+
+    return error.name === 'TimeoutError' || error.name === 'AbortError';
+}
+
+// Equal jitter (half fixed, half random) so the delay keeps a floor.
+function getRetryDelay(attempt: number, config: LongPollingConfig): number {
+    const backoffDelay = config.exponentialBackoff
+        ? Math.min(RETRY_BASE_DELAY_MS * 2 ** attempt, RETRY_MAX_DELAY_MS)
```
sdkennedy2: I think we want `attempt-1` don't we? That way attemp 1 will have a backoff of `RETRY_BASE_DELAY_MS` instead of `RETRY_BASE_DELAY_MS * 2`

```suggestion
        ? Math.min(RETRY_BASE_DELAY_MS * 2 ** (attempt - 1), RETRY_MAX_DELAY_MS)
```

sdkennedy2 (approved):

packages/plugins/apps/src/vite/dev-server.ts
Ayc0: `attempt` starts at 0 for the initial call, then 1 for the 1st retry etc.

I think the logic is fine as it is, no?

`delay * 2**retryCount`

And this is not called when attempt == 0:

Image

Ayc0: Edit: i'm wrong ^^

PR #502, opened 2026-09-03
[RUM-18236] Improve api for upload of source maps by debug ID

jhssilva:
### What and why?

Add direct debug-ID source-map uploads to Datadog Build Plugins. A build can now inject debug IDs and upload the corresponding source maps without configuring a service, release version, or minified path prefix.

The existing nested debug-ID and service/version workflows remain backward compatible.

- Documentation: https://github.com/DataDog/documentation/pull/39656

### How to use it

For Build Plugins to inject debug IDs and upload the source maps:

```js
datadogWebpackPlugin({
  auth: {
    apiKey: process.env.DATADOG_API_KEY,
    site: 'datadoghq.com',
  },
  sourcemaps: {
    debugId: true,
    upload: true,
  },
});
```

- `sourcemaps.debugId: true` injects a debug ID into each generated JavaScript bundle.
- `sourcemaps.upload: true` uploads the corresponding source maps during the build.
- Omitting `upload` enables injection without direct upload, for example when `datadog-ci` performs the upload.
- `auth.apiKey` or `DATADOG_API_KEY` is required only when direct upload is enabled.
- `bailOnError`, `dryRun`, and `maxConcurrency` can be configured alongside `upload: true`.

The existing `rum.sourceCodeContext` and `errorTracking.sourcemaps` configurations remain supported. They cannot be combined with the new top-level `sourcemaps` option.

### Existing service/version workflow

The legacy configuration continues to work unchanged:

```js
datadogWebpackPlugin({
  auth: {
    apiKey: process.env.DATADOG_API_KEY,
    site: 'datadoghq.com',
  },
  errorTracking: {
    sourcemaps: {
      service: 'my-application',
      releaseVersion: '1.0.0',
      minifiedPathPrefix: 'https://example.com/static/',
    },
  },
  rum: {
    sourceCodeContext: {
      service: 'my-application',
      version: '1.0.0',
    },
  },
});
```

### Implementation

- Add debug-ID upload support for webpack, Vite, Rollup, esbuild, and Rspack.
- Add a typed top-level `sourcemaps` option that distinguishes injection-only and upload-enabled configurations.
- Normalize the top-level option to the existing RUM injection and Error Tracking upload implementations.
- Represent upload behavior internally with the typed `SourcemapsUploadMode` discriminator.
- Build debug-ID multipart payloads without legacy service/version/path metadata.
- Require a debug ID for every artifact uploaded in debug-ID mode.
- Preserve the existing nested debug-ID and service/version behavior.

### Validation

- 77 focused unit and integration tests pass across the factory, RUM, and Error Tracking paths.
- Package typechecking, formatting, generated documentation/link checks, repository integrity, and open-source compliance pass locally.
- A real webpack smoke test using `sourcemaps: { debugId: true, upload: true }` uploaded a source map to the staging intake with `bailOnError: true`.

### DIFF
--- a/packages/factory/src/validate.ts
+++ b/packages/factory/src/validate.ts
@@ -10,6 +10,7 @@ import type {
     BuildMetadata,
     Options,
     OptionsWithDefaults,
+    SourcemapsOptions,
 } from '@dd/core/types';
 
 const SITES_DOC_URL = 'https://docs.datadoghq.com/getting_started/site/';
@@ -47,15 +48,93 @@ const validateMetadata = (metadata: BuildMetadata | undefined): string[] => {
     return errors;
 };
 
+const normalizeSourcemapsOptions = (options: Options, errors: string[]): Options => {
+    if (options.sourcemaps === undefined) {
+        return options;
+    }
+
+    if (
+        options.sourcemaps === null ||
+        typeof options.sourcemaps !== 'object' ||
+        Array.isArray(options.sourcemaps)
+    ) {
+        errors.push('sourcemaps must be an object');
+        return options;
+    }
+
+    const sourcemaps = options.sourcemaps as SourcemapsOptions;
+    const runtimeOptions = sourcemaps as unknown as Record<string, unknown>;
+
+    if (runtimeOptions.debugId !== true) {
+        errors.push('sourcemaps.debugId must be true');
+    }
+    if (runtimeOptions.upload !== undefined && typeof runtimeOptions.upload !== 'boolean') {
+        errors.push('sourcemaps.upload must be a boolean');
+    }
+    if (
+        runtimeOptions.upload !== true &&
+        ['bailOnError', 'dryRun', 'maxConcurrency'].some(
+            (option) => runtimeOptions[option] !== undefined,
+        )
+    ) {
+        errors.push(
+            'sourcemaps.bailOnError, sourcemaps.dryRun, and sourcemaps.maxConcurrency require sourcemaps.upload to be true',
+        );
+    }
+    if (options.rum?.sourceCodeContext !== undefined) {
+        errors.push('sourcemaps cannot be combined with rum.sourceCodeContext');
+    }
+    if (options.rum?.enable === false) {
+        errors.push('rum.enable cannot be false when sourcemaps is configured');
+    }
+    if (options.errorTracking?.sourcemaps !== undefined) {
+        errors.push('sourcemaps cannot be combined with errorTracking.sourcemaps');
+    }
+    if (runtimeOptions.upload === true && options.errorTracking?.enable === false) {
+        errors.push('errorTracking.enable cannot be false when sourcemaps.upload is true');
+    }
+
+    if (errors.length > 0) {
+        return options;
+    }
+
+    const normalized: Options = {
+        ...options,
+        rum: {
+            ...options.rum,
+            sourceCodeContext: { debugId: true },
+        },
+    };
+
+    if (sourcemaps.upload === true) {
+        normalized.errorTracking = {
+            ...options.errorTracking,
+            sourcemaps: {
+                debugId: true,
+                ...(sourcemaps.bailOnError !== undefined && {
+                    bailOnError: sourcemaps.bailOnError,
+                }),
+                ...(sourcemaps.dryRun !== undefined && { dryRun: sourcemaps.dryRun }),
+                ...(sourcemaps.maxConcurrency !== undefined && {
+                    maxConcurrency: sourcemaps.maxConcurrency,
+                }),
+            },
+        };
+    }
+
+    return normalized;
+};
+
 export const validateOptions = (options: Options = {}): OptionsWithDefaults => {
     const errors: string[] = validateMetadata(options.metadata);
+    const normalizedOptions = normalizeSourcemapsOptions(options, errors);
     // DATADOG_SITE env var takes precedence over configuration; only validate
     // auth.site when no env var is set, so a stale auth.site can't block a
     // build that has already opted into an env override.
     const envRaw = getDDEnvValue('SITE');
     const resolvedSite =
         resolveSite(envRaw, 'DATADOG_SITE/DD_SITE', errors) ??
-        resolveSite(options.auth?.site, 'auth.site', errors);
+        resolveSite(normalizedOptions.auth?.site, 'auth.site', errors);
 
     const auth: AuthOptionsWithDefaults = {
         site: resolvedSite?.site ?? DEFAULT_SITE,
@@ -68,12 +147,12 @@ export const validateOptions = (options: Options = {}): OptionsWithDefaults => {
 
     // Prevent these from being accidentally logged.
     Object.defineProperty(auth, 'apiKey', {
-        value: getDDEnvValue('API_KEY') || options.auth?.apiKey,
+        value: getDDEnvValue('API_KEY') || normalizedOptions.auth?.apiKey,
         enumerable: false,
     });
 
     Object.defineProperty(auth, 'appKey', {
-        value: getDDEnvValue('APP_KEY') || options.auth?.appKey,
+        value: getDDEnvValue('APP_KEY') || normalizedOptions.auth?.appKey,
         enumerable: false,
     });
 
@@ -81,7 +160,7 @@ export const validateOptions = (options: Options = {}): OptionsWithDefaults => {
         enableGit: true,
         logLevel: 'warn',
         metadata: {},
-        ...options,
+        ...normalizedOptions,
         auth,
     };
 };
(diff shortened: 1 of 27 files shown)


buranmert (commented):
i like the idea but i'm wondering [how can we make sure RUM and Error Tracking plugins are used together](https://datadoghq.atlassian.net/wiki/spaces/RUMP/pages/7167017802/Bundler+plugin+API?focusedCommentId=7178486130)? 
(i'm not familiar with how to use build plugins, i assume the user picks whichever plugins they want to use)

packages/plugins/error-tracking/src/types.ts
```diff
@@ -4,18 +4,55 @@
...
+export type ServiceVersionSourcemapsOptionsWithDefaults = SourcemapsUploadOptionsWithDefaults &
+    Required<
+        Pick<ServiceVersionSourcemapsOptions, 'minifiedPathPrefix' | 'releaseVersion' | 'service'>
+    > & {
+        mode: SourcemapsUploadMode.SERVICE_VERSION;
+    };
+
+export type DebugIdSourcemapsOptionsWithDefaults = SourcemapsUploadOptionsWithDefaults & {
+    mode: SourcemapsUploadMode.DEBUG_ID;
+};
```
amortemousque, line 46: **💬 suggestion:** ‏Why do we need a new type here? Instead of introducing a `mode` field, could we use `debugId` as the discriminant?


12:25 PM

packages/factory/src/validate.ts
```diff
@@ -47,18 +48,97 @@ const validateMetadata = (metadata: BuildMetadata | undefined): string[] => {
     return errors;
 };
 
+const normalizeSourcemapsOptions = (options: Options, errors: string[]): Options => {
```
amortemousque, line 51: **💬 suggestion:** ‏This function re-implements validation that rum/validate.ts and error-tracking/validate.ts already do (apiKey required, service/version conflicts, enable-flag conflicts), same rules, second place to keep in sync. Could we avoid repeating them

packages/plugins/error-tracking/src/validate.ts
```diff
@@ -62,59 +63,103 @@ export const validateSourcemapsOptions = (
...
+        toReturn.errors.push(
+            `${red('sourcemaps.minifiedPathPrefix')} must be a valid URL or start with '/'.`,
+        );
+    }
+
+    // Build the resolved config only when `releaseVersion` actually
+    // resolves; otherwise an error has been recorded and the caller will
+    // throw before the config is read.
+    if (releaseVersion) {
+        const { debugId: _debugId, ...serviceVersionOptions } = sourcemapsCfg;
```
amortemousque: **❓ question:** ‏Couldn't we directly pass `sourcemapsCfg` instead of adding this spread?

packages/plugins/error-tracking/src/validate.ts
```diff
@@ -62,59 +63,103 @@ export const validateSourcemapsOptions = (
...
+                `${red('auth.apiKey')} is required to upload source maps by debug ID.`,
             );
         }
 
-        // Build the resolved config only when `releaseVersion` actually
-        // resolves; otherwise an error has been recorded and the caller will
-        // throw before the config is read.
-        if (releaseVersion) {
+        if (toReturn.errors.length === 0) {
+            const { debugId: _debugId, ...uploadOptions } = sourcemapsCfg;
```
amortemousque: **❓ question:** ‏Couldn't we directly pass `sourcemapsCfg` instead of adding this spread?

packages/plugins/rum/src/types.ts
```diff
@@ -7,12 +7,22 @@ import type { Assign } from '@dd/core/types';
...
+    service?: never;
+    version?: never;
+};
+
+type ServiceVersionSourceCodeContextOptions = {
+    debugId?: false;
+    service: string;
     version?: string;
-    debugId?: boolean;
 };
```
amortemousque, line 20: **🔨 warning:** This assumption is incorrect. A customer can provide source code context with both a debugId and service/version. The service and version identify and allow to filter events from a specific micro-frontend; they are independent of the unminification.

packages/plugins/error-tracking/src/types.ts
jhssilva: Great suggestion.

jhssilva: Done in 4cbb00dd. I removed SourcemapsUploadMode and now use debugId as the discriminant in the normalized types and throughout file selection, payload creation, and upload metrics.

packages/factory/src/validate.ts
jhssilva: The normalizer is still needed to translate the new top-level sourcemaps option into the existing RUM and Error Tracking plugin configurations. It only validates the top-level shape and cross-plugin contradictions; API-key and service/version/path validation remain in the plugin validators. The explicit enable:false checks must happen here because disabled plugins do not run their validators. Let me know if you would prefer this split to be made more explicit in naming or structure.

packages/plugins/error-tracking/src/validate.ts
jhssilva: Yes, done in 4cbb00dd. The service/version branch now spreads sourcemapsCfg directly, then sets the normalized debugId:false value and resolved releaseVersion.

packages/plugins/error-tracking/src/validate.ts
jhssilva: Yes, done in 4cbb00dd. The debug-ID branch now spreads sourcemapsCfg directly, preserving debugId:true.

packages/plugins/rum/src/types.ts
jhssilva: Fixed in 4cbb00dd. The debug-ID source-code-context variant now accepts optional service and version, validation preserves them, and the injected context serializes service, version, and ddDebugId together. The legacy service/version variant remains unchanged.

packages/plugins/rum/src/getSourceCodeContextSnippet.ts
```diff
@@ -39,19 +39,19 @@ export const getSourceCodeContextSnippet = (
...
-    let debugId: string | undefined;
-    if (contextOptions.debugId) {
-        // Compute deterministic debug IDs whenever possible to prevent the backend from storing
-        // duplicate source maps for identical builds.
-        debugId = chunk ? stringToUUID(chunk.sourceOrHash) : randomUUID();
-    }
-
+    // Compute deterministic debug IDs whenever possible to prevent the backend from storing
+    // duplicate source maps for identical builds.
+    const debugId =
```
amortemousque: **🥜 nitpick:** Nested ternaries are difficult to read. Could we revert to the original code if the behavior is unchanged?

packages/plugins/error-tracking/src/sourcemaps/payload.ts
```diff
@@ -174,6 +175,9 @@ export const getPayload = async (
     if (!validity.sourcemap.exists) {
         errors.push(`Sourcemap file not found: ${sourcemap.sourcemapFilePath}`);
     }
+    if (debugIdRequired && !debugId) {
+        errors.push(`No debug ID found in minified file: ${sourcemap.minifiedFilePath}`);
```
amortemousque: **💬 suggestion:** ‏I don’t think we should abort all uploads when a single file is missing a debug ID. To remain consistent with `datadog-ci` upload: if all debug IDs are missing, fail with exit code 1; if only some are missing, skip those files and upload the rest.

packages/plugins/rum/src/getSourceCodeContextSnippet.ts
jhssilva: Agreed, the behavior is unchanged. I've reverted this to the original variable declaration and if-statement form to avoid the nested ternary. Thanks!

jhssilva: [c655c0c](https://github.com/DataDog/build-plugins/pull/502/commits/c655c0c476712a348dc06d4dbed96bfd79f9ca24)

packages/plugins/error-tracking/src/sourcemaps/payload.ts
jhssilva: Addressed [035b29f](https://github.com/DataDog/build-plugins/pull/502/commits/035b29fbf15cbe77812dac4bbca3b08654261cfd)

amortemousque (approved):

packages/plugins/error-tracking/src/sourcemaps/sender.ts
amortemousque, line 1: Not fully fan with the validation flow in this file, but it could be improve in followup PRs

amortemousque (approved):

nchapma2 (approved):
Approving for build-plugins. I don't really know this code at all and in the plugin files, it seems like we could do some cleaning up if debug-ids are the identifier going forward

PR #510, opened 2026-09-09
[APPS-2792] Fix: harden env-guard against bypass and reliability gaps

tyffical:
## Motivation

- Tracked in [APPS-2792](https://datadoghq.atlassian.net/browse/APPS-2792) (Milestone: Secret Store parity for local execution). Stacked on #504, which introduces `env-guard.ts`'s core `process.env` scoping mechanism.
- Repeated review of that mechanism surfaced further bypass and reliability gaps in the guarded `fs`/`process.report`/`util.inspect`/Proxy surface. See the Changes table below for the full, itemized list.

## Architecture

All fixes land inside `env-guard.ts`'s existing scoping mechanism (introduced in #504) — this PR closes bypass and reliability gaps in that mechanism rather than changing its shape:

```
runWithScopedEnv(scopedEnv, fn)
  │
  ├─ fs wrappers — coverage extended to fs.read/readv(Sync)/openAsBlob/
  │    FileReadStream and every FileHandle.prototype read method, with a
  │    real fd resolved via a captured native getter so an own-property
  │    shadow on the handle can't fool the guard; cp/cpSync/promises.cp
  │    reject recursive+dereference outright and read their options once
  │
  ├─ process.env — reassigning to null/undefined/a primitive is rejected;
  │    Object.defineProperty(..., { configurable: false }) throws a clear
  │    guard-specific error instead of a native Proxy invariant violation
  │
  ├─ process.report — writeReport() refuses a non-regular destination,
  │    re-checked immediately before the real write to narrow a symlink-
  │    swap TOCTOU window; a deferred excludeEnv write is now validated
  │    via Node's native setter
  │
  └─ util.inspect(process.env) — Proxy target repointed to a permanently
       empty dummy object, closing the default-mode + showProxy:true leak
```

<details><summary>16 changes across env-guard.ts, env-guard.test.ts, guarded-wrapper.ts</summary>

| What changed | File |
|---|---|
| `fs.read`/`readSync`/`readv`/`readvSync` and every `FileHandle.prototype` read method (`read`/`readFile`/`readv`/`createReadStream`/`readableWebStream`/`readLines`) are now guarded against the `/proc/.../environ` bypass | `packages/plugins/apps/src/vite/env-guard.ts` |
| The `FileHandle` guard resolves the real fd via a getter captured once from the prototype at patch time, so a customer-controlled own property shadowing `.fd` with a harmless value can't fool the check | `packages/plugins/apps/src/vite/env-guard.ts` |
| That fd resolution — and the guard check depending on it — now only runs inside an active scope; the native `.fd` getter (and the guard's own logic) was previously invoked unconditionally on every call | `packages/plugins/apps/src/vite/env-guard.ts` |
| `fs.cp`/`cpSync`/`promises.cp` now reject a `recursive: true, dereference: true` copy outright, since Node's own internal traversal for that case bypasses the top-level source-path check | `packages/plugins/apps/src/vite/env-guard.ts` |
| `guardCpOptions`/`guardEnvironPathOrFdOption` read a getter-backed option value exactly once via destructuring, instead of risking a second, differently-answered read during a later spread | `packages/plugins/apps/src/vite/env-guard.ts` |
| `fs.openAsBlob` and `fs.FileReadStream` are now guarded, each gated on the real function actually existing on the running Node version | `packages/plugins/apps/src/vite/env-guard.ts` |
| Reassigning `process.env` to `null`, `undefined`, or a primitive (not an object) is now rejected instead of silently corrupting the real environment fallback | `packages/plugins/apps/src/vite/env-guard.ts` |
| `Object.defineProperty(process.env, key, { configurable: false })` now throws a clear, guard-specific error instead of a native Proxy invariant `TypeError` — a non-configurable definition can never be satisfied against the Proxy's permanently-empty target | `packages/plugins/apps/src/vite/env-guard.ts` |
| `process.env`'s Proxy now targets a permanently-empty dummy object, closing `util.inspect(process.env)` reading the real environment straight off the Proxy's internal target — both default mode and `showProxy: true` bypass every trap by design, per Node's own docs | `packages/plugins/apps/src/vite/env-guard.ts` |
| `process.report.writeReport()` refuses a verified non-regular destination (FIFO/socket/character device), re-checked immediately before the real write (not just once, earlier in the function) to narrow the window for an external symlink swap | `packages/plugins/apps/src/vite/env-guard.ts` |
| A deferred `excludeEnv` write (made from outside an active scope while a different scope is active) is now validated immediately via Node's native setter instead of skipping that validation | `packages/plugins/apps/src/vite/env-guard.ts` |
| `wrapGuardedAsyncFsFn`'s `onResolved` hook (used by `fs.promises.open`) now forwards the caller's `this` through to the wrapped function, instead of losing it on a bare call | `packages/plugins/apps/src/vite/env-guard.ts` |
| Two inlined function-call arguments (a repo convention violation) replaced with named locals | `packages/plugins/apps/src/vite/env-guard.ts` |
| New tests for every fix above: FileHandle fd-shadow bypass, fd-based reads, `openAsBlob`/`FileReadStream` guards, cp TOCTOU, primitive `process.env` rejection, the `defineProperty` invariant, and the existing `/proc/.../environ`/scope tests extended to the newly-guarded entry points | `packages/plugins/apps/src/vite/env-guard.test.ts` |
| `makeGuardWrapper`/`makeGuardCallbackWrapper` doc comment tightened | `packages/plugins/apps/src/vite/guarded-wrapper.ts` |

</details>

## QA Instructions

```bash
yarn typecheck:all
# Expected: no errors ✅ VERIFIED

yarn build:all
yarn test:unit
# Expected: 94 suites, 2405 tests (2404 passed, 1 pre-existing unrelated skip) ✅ VERIFIED
```

<details><summary>Manual QA: standalone-process reproduction for the util.inspect(process.env) fix (click to expand — the only fix here that needs a genuinely fresh OS process, not just an in-process Jest test)</summary>

Every other fix in this PR is exercised directly by a named test in `env-guard.test.ts` (run `yarn test:unit packages/plugins/apps/src/vite/env-guard.test.ts` to reproduce all of them). The `util.inspect` fix needs a fresh, never-before-scoped `process.env` to observe the Proxy's real target, which a single Jest test file (already sharing the module's installed state across its own tests) can't produce on its own — verified instead via a standalone script:

```ts
// /tmp/inspect-proxy-repro.mjs
import util from 'util';

process.env.REAL_SECRET_ONLY_HERE = 'sk_should_never_leak_via_inspect';

const { runWithScopedEnv } = await import(
    '/path/to/build-plugins/packages/plugins/apps/src/vite/env-guard.ts'
);

await runWithScopedEnv({ PATH: '/scoped' }, async () => {
    const withoutShowProxy = util.inspect(process.env);
    const withShowProxy = util.inspect(process.env, { showProxy: true });
    const leaked =
        withoutShowProxy.includes('sk_should_never_leak_via_inspect') ||
        withShowProxy.includes('sk_should_never_leak_via_inspect');
    console.log(leaked ? 'LEAKED' : 'NOT LEAKED');
});
```

```bash
npx tsx /tmp/inspect-proxy-repro.mjs
# NOT LEAKED ✅ VERIFIED
```

</details>

<details><summary>Manual QA: env-guard's process.env scoping under a real end-to-end $.Actions call (click to expand — exercises the guard inside the actual dev-server request path, not just Jest)</summary>

- The tests above exercise env-guard in isolation. This reproduction drives it through the real path a customer function takes: `POST /__dd/executeAction` → staging auth → runtime-context priming (`preview-async`, long-polled) → in-process execution of the function body under `runWithScopedEnv` → a live `$.Actions` proxy call.
- Reproduced against a scaffolded consumer app with `@datadog/vite-plugin` built from this branch (`yarn build` in `packages/published/vite-plugin`) and imported by absolute path in the app's `vite.config.ts`, since `publishConfig`-based `exports` only resolve via a real `npm publish`, not a local link.

```ts
// src/envGuardActionsCheck.backend.ts
export async function envGuardActionsCheck() {
    const results: Record<string, unknown> = {};
    results.pathVisible = process.env.PATH !== undefined;
    results.realSecretHidden = process.env.QA_REAL_SECRET_MARKER === undefined;
    try {
        Object.defineProperty(process.env, 'LOCKED', { value: 'x', configurable: false });
        results.defineNonConfigurable = 'did not throw (unexpected)';
    } catch (error) {
        results.defineNonConfigurable = (error as Error).message;
    }
    try {
        const actionResult = await $.Actions.slack.chat.postMessage({
            inputs: { channel: '#test', text: 'env-guard QA' },
            connectionId: 'qa-connection-id',
        });
        results.actionsCallSucceeded = true;
        results.actionsCallResult = actionResult;
    } catch (error) {
        results.actionsCallSucceeded = false;
        results.actionsCallError = (error as Error).message;
    }
    return results;
}
```

```bash
# Import the function above from App.tsx, then start the dev server with real staging credentials
# (site set to dd.datad0g.com in vite.config.ts's auth block):
dd-auth --domain dd.datad0g.com -- sh -c 'npm run dev'

# Load the app once in a browser so Vite's transform hook registers the function, then call it:
HASH=$(node -e "console.log(require('crypto').createHash('sha256').update('src/envGuardActionsCheck').digest('hex'))")
curl -s -X POST http://localhost:5173/__dd/executeAction \
  -H 'Content-Type: application/json' \
  -d "{\"functionName\": \"${HASH}.envGuardActionsCheck\", \"args\": []}"
# Expected: pathVisible: true, realSecretHidden: true, defineNonConfigurable: guard-specific
# error message (not a native Proxy invariant TypeError), actionsCallSucceeded: false,
# actionsCallError: "Action $.Actions.slack.chat.postMessage used connection \"qa-connection-id\",
# which is not in this function's allowed connections: []" ✅ VERIFIED
```

- The connection-allowlist rejection at the end confirms the call reached PR #479's real enforcement layer — i.e. every layer in front of it (staging auth, runtime-context priming, in-process execution under this PR's env-guard fixes) ran for real rather than being mocked or short-circuited.

</details>

## Blast Radius

- Defense-in-depth hardening of an already-shipped scoping mechanism (#504) — no production-facing behavior change outside local dev-server execution.
- Risk: low. These are bypass-closing and reliability fixes to a JS-level guard already scoped to the dev server's local-execution path, not a hard security boundary.

## Out of Scope / Follow-ups

<details><summary>1 item deferred</summary>

| Item | Status | Next step |
|---|---|---|
| `process.report.writeReport()`'s TOCTOU window is narrowed (re-checked immediately before the real write) but not fully closed — a symlink swap in the remaining gap between that check and Node's own write syscall is still theoretically possible | Deferred, accepted | No action planned: exploiting this requires the backend function itself (or a coordinated external process) to actively race its own call, which is outside this guard's stated "JS-level defense-in-depth, not a hard security boundary" threat model |

</details>

## Documentation

- [APPS-2792](https://datadoghq.atlassian.net/browse/APPS-2792)
- [RFC (Confluence)](https://datadoghq.atlassian.net/wiki/spaces/HCA/pages/7012712651) — architecture, Decisions and Trade-Offs, Security, Open Questions
- [Kickoff (Confluence)](https://datadoghq.atlassian.net/wiki/spaces/HCA/pages/7025394455) — milestone breakdown, dependency graph, status

[APPS-2792]: https://datadoghq.atlassian.net/browse/APPS-2792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APPS-2792]: https://datadoghq.atlassian.net/browse/APPS-2792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/plugins/apps/src/vite/env-guard.ts
+++ b/packages/plugins/apps/src/vite/env-guard.ts
@@ -493,6 +493,35 @@ function extractFdNumber(fdValue: unknown): unknown {
     return fdValue;
 }
 
+// Reads each of `keys` from a getter-backed options object exactly once, then rebuilds a plain
+// object where a key absent from the caller's own options stays absent — see guardCpOptions's own
+// comment for why re-adding an absent key as explicit `undefined` breaks Node's own cp validation.
+// Shared by guardEnvironPathOrFdOption (a single always-present key) and guardCpOptions (two
+// independently-optional keys): presence-conditional restore is correct for both, since a
+// guaranteed-present key just never hits the "absent" branch.
+function snapshotOptionKeysOnce<K extends string>(
+    options: Record<string, unknown>,
+    keys: readonly K[],
+): Record<string, unknown> {
+    // Copying every OTHER own-enumerable key first, via Object.keys rather than a `{ ...options }`
+    // spread, is what keeps this to exactly one read per snapshotted key below — a spread of the
+    // full options object would invoke every key's getter once on its own, then a second time when
+    // that same key is explicitly snapshotted next.
+    const keySet = new Set<string>(keys);
+    const result: Record<string, unknown> = {};
+    for (const key of Object.keys(options)) {
+        if (!keySet.has(key)) {
+            result[key] = options[key];
+        }
+    }
+    for (const key of keys) {
+        if (key in options) {
+            result[key] = options[key];
+        }
+    }
+    return result;
+}
+
 // createReadStream/ReadStream's options.fd (a raw fd number, or a FileHandle whose own .fd is one)
 // makes Node read from that fd directly, ignoring the leading path argument — a plain
 // throwIfBlockedEnvironPath(rawPath) would never see the real target. Returns a safe options
@@ -510,12 +539,9 @@ function guardEnvironPathOrFdOption(rawPath: unknown, options: unknown): unknown
     if (typeof options !== 'object' || options === null || !('fd' in options)) {
         return options;
     }
-    // Destructuring reads the getter exactly once, into fdValue — spreading the remainder (with fd
-    // already removed) can't invoke it again the way `{ ...options, fd: fdValue }` would have.
-    const { fd: fdValue, ...restOptions } = options as { fd: unknown };
-    const fdNumber = extractFdNumber(fdValue);
-    throwIfBlockedEnvironPath(fdNumber);
-    return { ...restOptions, fd: fdValue };
+    const snapshot = snapshotOptionKeysOnce(options as Record<string, unknown>, ['fd'] as const);
+    throwIfBlockedEnvironPath(extractFdNumber(snapshot.fd));
+    return snapshot;
 }
 
 // Every guarded fs entry point below except createReadStream takes only a leading path argument —
@@ -778,26 +804,18 @@ function guardCpOptions(options: unknown): unknown {
     if (typeof options !== 'object' || options === null) {
         return options;
     }
-    const {
-        recursive: recursiveValue,
-        dereference: dereferenceValue,
-        ...restOptions
-    } = options as { recursive?: unknown; dereference?: unknown };
-    if (recursiveValue === true && dereferenceValue === true) {
+    // Node's cp implementations distinguish an absent key (defaulted internally) from one
+    // explicitly present with value `undefined` (rejected by its own validation), so
+    // snapshotOptionKeysOnce restoring both keys unconditionally would turn a caller's
+    // `{ recursive: true }` (no dereference key at all) into `{ recursive: true, dereference:
+    // undefined }` and throw a validation error that never happens when the real options object is
+    // forwarded as-is — this is why it only re-adds keys actually present on the caller's options.
+    const safeOptions = snapshotOptionKeysOnce(
+        options as Record<string, unknown>,
+        ['recursive', 'dereference'] as const,
+    );
+    if (safeOptions.recursive === true && safeOptions.dereference === true) {
         throw new Error(CP_BLOCKED_MESSAGE);
-    }
-    // Only re-adds a key that was actually present on the caller's own options — Node's cp
-    // implementations distinguish an absent key (defaulted internally) from one explicitly present
-    // with value `undefined` (rejected by its own validation), so restoring both keys unconditionally
-    // would turn a caller's `{ recursive: true }` (no dereference key at all) into
-    // `{ recursive: true, dereference: undefined }` and throw a validation error that never happens
-    // when the real options object is forwarded as-is.
-    const safeOptions = restOptions as Record<string, unknown>;
-    if ('recursive' in options) {
-        safeOptions.recursive = recursiveValue;
-    }
-    if ('dereference' in options) {
-        safeOptions.dereference = dereferenceValue;
     }
     return safeOptions;
 }
@@ -813,11 +831,8 @@ const originalPromisesCp = fs.promises.cp;
 // implementations don't consult it, but nothing here should be the one silent exception.
 fs.cpSync = function (this: unknown, src: unknown, dest: unknown, options?: unknown) {
     throwIfBlockedEnvironPath(src);
-    return originalCpSync.apply(this, [
-        src as string | URL,
-        dest as string | URL,
-        guardCpOptions(options) as fs.CopySyncOptions,
-    ]);
+    const safeOptions = guardCpOptions(options) as fs.CopySyncOptions;
+    return originalCpSync.apply(this, [src as string | URL, dest as string | URL, safeOptions]);
 } as typeof fs.cpSync;
 
 // cp reports failure via an error-first callback, never a synchronous throw. Only the guard's own
@@ -844,11 +859,8 @@ fs.cp = function (this: unknown, src: unknown, dest: unknown, ...rest: unknown[]
 // promises.cp rejects, matching its real Promise-returning contract.
 fs.promises.cp = async function (this: unknown, src: unknown, dest: unknown, options?: unknown) {
     throwIfBlockedEnvironPath(src);
-    return originalPromisesCp.apply(this, [
-        src as string | URL,
-        dest as string | URL,
-        guardCpOptions(options) as fs.CopyOptions,
-    ]);
+    const safeOptions = guardCpOptions(options) as fs.CopyOptions;
+    return originalPromisesCp.apply(this, [src as string | URL, dest as string | URL, safeOptions]);
 } as typeof fs.promises.cp;
 
 // createReadStream's own wrap above only covers that factory function — Node also exports the
--- a/packages/plugins/apps/src/vite/env-guard.ts
+++ b/packages/plugins/apps/src/vite/env-guard.ts
@@ -10,7 +10,7 @@ import { syncBuiltinESMExports } from 'node:module';
 import nodePath from 'path';
 import { fileURLToPath } from 'url';
 
-import { makeGuardCallbackWrapper, makeGuardWrapper } from './guarded-wrapper';
+import { invokeCallbackArg, makeGuardCallbackWrapper, makeGuardWrapper } from './guarded-wrapper';
 import { getOrCreateShared } from './shared-module-singleton';
 
 // Captured at module load — before any customer code runs — so a backend function can't replace
@@ -185,7 +185,13 @@ function getSharedState(): EnvGuardSharedState {
                 if (activeScopeTokens.size > 0) {
                     // An unrelated caller writing from outside any scope while a DIFFERENT scope is
                     // still active elsewhere — applying it immediately would disarm redaction out
-                    // from under that scope, so it's deferred until the active scope's own cleanup.
+                    // from under that scope, so it's deferred instead. Applied and immediately
+                    // reverted here (rather than just stashed) so Node's own setter validation still
+                    // runs now — an invalid value throws here instead of surfacing later,
+                    // misattributed to whichever scope's cleanup happens to apply it.
+                    const currentValue = getExcludeEnv();
+                    applyExcludeEnvValue(newValue);
+                    applyExcludeEnvValue(currentValue);
                     savedExcludeEnv = newValue;
                     return;
                 }
@@ -200,6 +206,15 @@ function getSharedState(): EnvGuardSharedState {
                 if (nativeGetStore() !== undefined) {
                     throw new Error(
                         "Reassigning process.env is not allowed in backend functions — it would corrupt the dev server's real environment for every future execution. Use $.Source or a declared Custom Credential instead.",
+                    );
+                }
+                if (typeof newValue !== 'object' || newValue === null) {
+                    // isEnvProxy() only ever returns true for an object, so without this check a
+                    // primitive assignment (process.env = 1, process.env = null) would store that
+                    // primitive as realEnv — every later unscoped access then calls
+                    // Reflect.get/ownKeys on it and throws, permanently breaking process.env.
+                    throw new Error(
+                        'Reassigning process.env to a non-object value is not allowed — it would permanently break every future process.env access.',
                     );
                 }
                 realEnvHistory.push(realEnv);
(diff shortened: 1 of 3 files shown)


tyffical:
@codex review

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -403,6 +462,93 @@ function wrapGuardedStreamFn<T extends (...args: never[]) => unknown>(real: T):
...
+function patchFileHandleAsyncMethod(proto: Record<PropertyKey, unknown>, methodName: string): void {
+    const original = originalIfUnguarded(proto, methodName);
+    if (!original) {
+        return;
+    }
+    const guarded = function (this: { fd: number }, ...args: never[]): unknown {
+        let blocked: boolean;
+        try {
+            blocked = isBlockedEnvironPath(this.fd);
+        } catch (error) {
```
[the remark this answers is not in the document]
tyffical: Fixed: captured FileHandle.prototype's native `fd` getter once, at patch-install time (from the first real handle `fs.promises.open()` returns), and every guarded method now invokes it directly via `.call(handle)` instead of reading `this.fd` through ordinary property access — bypassing any later own-property shadow. New test verified on any OS (not just Linux) via the same mocked-`readlinkSync` pattern used elsewhere in this file: opens a real handle to a safe file, mocks its real fd as resolving to environ, then shadows `handle.fd` with a harmless value and confirms the read is still blocked.

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -411,17 +557,74 @@ fs.promises.readFile = wrapGuardedAsyncFsFn(fs.promises.readFile);
...
+        return false;
+    }
+    if (typeof options !== 'object' || options === null) {
+        return false;
+    }
+    return (
+        'recursive' in options &&
+        options.recursive === true &&
+        'dereference' in options &&
+        options.dereference === true
```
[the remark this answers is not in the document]
tyffical: Fixed: cpSync/cp/promises.cp are now hand-rolled wrappers (bypassing the generic makeGuardWrapper machinery, the same way wrapGuardedStreamFn already does for createReadStream) that snapshot `options.recursive`/`options.dereference` into plain data properties before forwarding — the real implementation can only ever see the exact values already checked, never a getter's later, different return value. New test confirms both properties are read exactly once each, not read again by the real cp call.

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -217,6 +244,15 @@ function ensureEnvProxyInstalled(): void {
...
             );
+            if (newValue === null || newValue === undefined) {
+                // isEnvProxy(null|undefined) is false, so without this check the assignment below
+                // would store null/undefined as realEnv, permanently breaking every later unscoped
+                // process.env access — there's no legitimate reason to replace the whole object with
+                // either.
+                throw new Error(
+                    'Reassigning process.env to null or undefined is not allowed — it would permanently break every future process.env access.',
+                );
+            }
```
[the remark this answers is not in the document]
tyffical: Fixed: the check is now `typeof newValue !== 'object' || newValue === null`, rejecting every primitive (including null/undefined, since `typeof undefined !== 'object'`) in one pass instead of special-casing just the two nullish values. New test confirms reassigning to a number throws and doesn't corrupt the real environment.

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -411,17 +557,74 @@ fs.promises.readFile = wrapGuardedAsyncFsFn(fs.promises.readFile);
...
-fs.promises.open = wrapGuardedAsyncFsFn(fs.promises.open);
-
-// copyFileSync/copyFile/promises.copyFile/cpSync/promises.cp read the source file's bytes through
-// a distinct native binding that never calls through readFile*/open* above — an uncovered path that
-// could otherwise copy /proc/.../environ to an ordinary, unguarded file and read it back from there.
+fs.promises.open = wrapGuardedAsyncFsFn(fs.promises.open, ensureFileHandleReadGuarded);
+
+// openAsBlob resolves a path straight to a Blob's lazily-read bytes, its own entry point separate
+// from open*/readFile* above.
+fs.openAsBlob = wrapGuardedAsyncFsFn(fs.openAsBlob);
```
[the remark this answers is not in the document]
tyffical: Fixed: the assignment is now gated on `typeof fs.openAsBlob === 'function'`, matching the existing feature-detection convention in packages/core/src/helpers/fs.ts's own getFile(). New test simulates Node 18 by deleting fs.openAsBlob and re-evaluating the module via jest.isolateModules(), confirming it stays undefined rather than becoming a wrapper around undefined.

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -693,15 +735,121 @@ fs.openSync = wrapGuardedFsFn(fs.openSync);
...
+// Captured into local consts before reassignment below — a lazy `() => fs.cpSync` getter would
+// re-read the property AFTER it's replaced with this very wrapper, recursing into itself forever.
+const originalCpSync = fs.cpSync;
+const originalCp = fs.cp;
+const originalPromisesCp = fs.promises.cp;
+
+// cpSync is genuinely synchronous — a guard failure throwing matches its real contract. `this`
+// is forwarded via .apply, matching every other guarded fs entry point in this file — Node's own
+// implementations don't consult it, but nothing here should be the one silent exception.
+fs.cpSync = function (this: unknown, src: unknown, dest: unknown, options?: unknown) {
```
ksun154, line 832: guardCpOptions(options) is passed inline into .apply(...) here, which is the same "inlined function-call argument" pattern this PR's description says was fixed elsewhere. The sibling fs.cp callback wrapper assigns it to safeOptions first. It's probably worth doing the same in cpSync/promises.cp for consistency.

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -693,15 +735,121 @@ fs.openSync = wrapGuardedFsFn(fs.openSync);
...
+            [src, dest, ...rest],
+            error instanceof Error ? error : new Error(String(error)),
+        );
+        return undefined;
+    }
+    return (originalCp as unknown as (...a: unknown[]) => unknown).apply(this, safeArgs);
+} as typeof fs.cp;
+
+// promises.cp rejects, matching its real Promise-returning contract.
+fs.promises.cp = async function (this: unknown, src: unknown, dest: unknown, options?: unknown) {
```
ksun154, line 860: same inlined function-call argument

packages/plugins/apps/src/vite/env-guard.ts
```diff
@@ -693,15 +735,121 @@ fs.openSync = wrapGuardedFsFn(fs.openSync);
...
+    "Copying /proc/.../environ, or recursively copying with dereference: true, is not allowed in backend functions — both can expose the dev server's real, unscoped environment. Use $.Source or a declared Custom Credential instead.";
+
+// Snapshots options.recursive/dereference into plain data properties, read exactly once — the
+// generic makeGuardWrapper machinery forwards a caller's original options object unchanged, which
+// would let a getter-backed recursive/dereference report false here and true when Node's own
+// fs.cp* reads it again internally, letting a symlinked /proc/.../environ get dereferenced and
+// copied through undetected. Same TOCTOU reasoning as guardEnvironPathOrFdOption's options.fd
+// handling above; hand-rolled here since arg transformation isn't something makeGuardWrapper
+// supports.
+function guardCpOptions(options: unknown): unknown {
```
ksun154, line 800: guardCpOptions and guardEnvironPathOrFdOption both reimplement "snapshot a getter-backed option once, then conditionally rebuild" with subtly different restore semantics (unconditional vs. presence-conditional). A shared snapshotOptionKeysOnce helper would prevent the two from drifting apart on a future fix.

packages/plugins/apps/src/vite/env-guard.ts
ksun154: also nit:  cpSync/cp/promises.cp don't use the shared makeGuardWrapper/makeGuardCallbackWrapper/wrapGuardedAsyncFsFn helpers. Each one re-implements its own guard logic instead. Not blocking, but it means any future fix to how guards dispatch (like the this-forwarding fix in this PR) has to be applied in three separate places instead of one. Might be worth reusing the shared wrapper here if that's feasible.

tyffical: Fixed: cpSync and promises.cp now assign guardCpOptions(options) to a local safeOptions before forwarding it, matching the plain cp wrapper.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/plugins/apps/src/vite/env-guard.ts
tyffical: Fixed alongside the cpSync instance — promises.cp now assigns to a local safeOptions too.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

packages/plugins/apps/src/vite/env-guard.ts
tyffical: Fixed: extracted a shared snapshotOptionKeysOnce helper — presence-conditional restore is correct for both guardCpOptions and guardEnvironPathOrFdOption, since a guaranteed-present key just never hits the absent branch.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

ksun154 (approved):

PR #515, opened 2026-09-17
fix(live-debugger): pass an invocation handle from entry to exit hooks

watson:
### What and why?

Instrumented functions kept in-flight probe state in a LIFO stack shared by every invocation, and the exit hooks popped the most recent entry because nothing told them which invocation was exiting. Overlapping async calls that finish out of entry order get paired with the wrong entry — wrong duration, and entry arguments from one call attached to another call's return value or exception. An invocation that never settles is worse: its entry is never popped, so the stack stays offset for every later call.

`$dd_entry` now returns a handle identifying the invocation, and every exit path hands it back.

### How?

The generated code already called `$dd_probes` once per invocation and bound the result to a function-scoped variable, so the handle reuses that binding instead of adding a new variable or a new hook parameter. Two template strings change; the catch block and all four exit forms are untouched.

```diff
-const probeDecl = `const ${probeVarName} = $dd_probes('${escapedFunctionId}');`;
-const entryCall = `if (${probeVarName}) $dd_entry(${probeVarName}, ${receiverArg}${argsArg});`;
+const probeDecl = `let ${probeVarName} = $dd_probes('${escapedFunctionId}');`;
+const entryCall = `if (${probeVarName}) ${probeVarName} = $dd_entry(${probeVarName}, ${receiverArg}${argsArg});`;
```

Generated output, before:

```js
function add(a, b) {
    const $dd_p0 = $dd_probes('src/utils.js;add');
    try {
        let $dd_rv0;
        if ($dd_p0) $dd_entry($dd_p0, this, {a, b});
        const sum = a + b;
        return ($dd_rv0 = sum, $dd_p0 ? $dd_return($dd_p0, $dd_rv0, this, {a, b}, {sum}) : $dd_rv0);
    } catch(e) { if ($dd_p0) $dd_throw($dd_p0, e, this, {a, b}); throw e; }
}
```

After:

```js
function add(a, b) {
    let $dd_p0 = $dd_probes('src/utils.js;add');
    try {
        let $dd_rv0;
        if ($dd_p0) $dd_p0 = $dd_entry($dd_p0, this, {a, b});
        const sum = a + b;
        return ($dd_rv0 = sum, $dd_p0 ? $dd_return($dd_p0, $dd_rv0, this, {a, b}, {sum}) : $dd_rv0);
    } catch(e) { if ($dd_p0) $dd_throw($dd_p0, e, this, {a, b}); throw e; }
}
```

Hook signatures are unchanged — `$dd_entry`'s return type goes from `void` to a handle, and the exit hooks' first parameter is reinterpreted. Since the handle is `undefined` when no probe produced an entry (sampling, entry conditions, lifetime budgets), the exit guards go false and the return and throw hooks are now skipped entirely on that path instead of being called to do nothing.

`invocationHandle.test.ts` is new: it runs the transformed code in a `vm` sandbox against a stub runtime, which nothing in this plugin did before. It asserts the exact object `$dd_entry` returned reaches the exit hook across all eight exit forms, that overlapping async calls pair correctly settling in both orders and when one throws, and that a skipped entry still returns the right value to the caller.

Docs updated to match: README gains an "Invocation handles" section, and the example catalog in EXAMPLES.md is regenerated.

### Note for the reviewer

This is one half of the fix. `onEntry` in `browser-sdk` still returns `void`, so until the matching change lands there `$dd_p0` is `undefined` after entry and no snapshots are emitted. The SDK side needs to return a list of `{ probe, state }` pairs captured at entry, iterate that in `onReturn`/`onThrow` instead of `probe.activeEntries.pop()`, drop `InitializedProbe.activeEntries`, and make `onEntry` non-throwing.


P403n1x87 (approved):

yoannmoinet (approved):
I'm not really knowledgeable with this code.
Mostly rubber stamping to unlock the PR.

PR #517, opened 2026-09-18
Fix flaky apps dev-server timeout tests

ksun154:
## What does this PR do?

- Gives the priming-timeout test its own permanently pending `loadModule` mock.
- Always releases the concurrent priming gate and drains both responses, even when the assertion fails.

## Why?

A timed-out `loadCustomerModuleEntry()` is abandoned rather than cancelled. The test previously passed the shared mutable `mockLoadModule`, so a late continuation could run after Jest reset that mock for the next test. That produced the duplicate `greet-priming-start` seen in [the CI failure](https://github.com/DataDog/build-plugins/actions/runs/35349960578/job/105615463955?pr=516), then leaked queued `compute` work into the following test.

## Verification

- `yarn test:unit packages/plugins/apps/src/vite/dev-server.test.ts --runInBand` — 51/51 passed
- Affected three-test sequence repeated 20 times — 0 failures
- `yarn workspace @dd/apps-plugin typecheck`
- `yarn eslint packages/plugins/apps/src/vite/dev-server.test.ts --quiet`


yoannmoinet (approved):
Thanks for this!

